PR #100, opened 2014-12-17
Fix CFY-1828:  cfy fails if .cloudify/config.yaml is missing

iconoeugen:
Fix is skipping the logger initialization if the .cloudify/config.yaml file is missing to aviod failing before executing command.


barakm:
Outdated, closing

PR #106, opened 2015-01-04
parsing input as a string dict

noakup:


cloudify_cli/utils.py
```diff
@@ -122,17 +122,53 @@ def inputs_to_dict(resource, resource_name):
...
+    else:
+        msg = "Invalid input: {0}. {1} must be either be a path to a valid " \
+              "YAML file, a string formatted as a valid YAML or a string " \
+              "formatted as a dictionary (key1=value1;key2=value2)"\
+            .format(resource, resource_name)
         raise CloudifyCliError(msg)
 
 
+def plain_string_to_dict(input_string):
+    input_string = str(input_string).strip()
```
dankilman: no need to call str(input_string). it could simply be input_string.strip(). If someone passes a non string to this function, it's their bug.

cloudify_cli/utils.py
```diff
@@ -122,17 +122,53 @@ def inputs_to_dict(resource, resource_name):
...
+    input_dict = {}
+    mapped_inputs = input_string.split(';')
+    for mapped_input in mapped_inputs:
+        mapped_input = str(mapped_input).strip()
+        if len(mapped_input) == 0:
```
dankilman: empty string have a boolean value of false.
because you know mapped input is a string you can simply do
`if not mapped_input` instead of `if len(mapped_input) == 0`

cloudify_cli/utils.py
```diff
@@ -122,17 +122,53 @@ def inputs_to_dict(resource, resource_name):
...
+                parsed_dict = yaml.load(resource)
+        except yaml.error.YAMLError as e:
+            msg = ("'{0}' is not a valid YAML. {1}"
+                   .format(resource_name, str(e)))
+            raise CloudifyCliError(msg)
+
+    if isinstance(parsed_dict, dict):
+        return parsed_dict
+    else:
+        msg = "Invalid input: {0}. {1} must be either be a path to a valid " \
```
dankilman: I think the message here is a bit misleading.
If the key=value syntax is applied, you are guaranteed to get a valid dictionary.
If on the other hand I supply a YAML string or YAML file that don't represent a dict, this is where the trouble lies.
So the message should be clear that the YAML variations should represent a valid dict/hash/map/object

cloudify_cli/utils.py
```diff
@@ -122,17 +122,53 @@ def inputs_to_dict(resource, resource_name):
...
-        msg = ("'{0}' must be a valid YAML. {1}"
-               .format(resource_name, str(e)))
+        # parse resource as string representation of a dictionary
+        parsed_dict = plain_string_to_dict(resource)
+    except CloudifyCliError:
+        try:
+            # if resource is a path - parse as a yaml file
+            if os.path.exists(resource):
+                with open(resource, 'r') as f:
+                    return yaml.load(f.read())
```
dankilman: `return yaml.load(f.read())` should be `parsed_dict = yaml.load(f.read())`

cloudify_cli/tests/test_utils.py
```diff
@@ -102,3 +105,92 @@ def test_dump_cosmo_working_dir_settings_create(self):
...
+                                utils.plain_string_to_dict,
+                                input_str)
+
+        input_str = "my_key1:my_value1;my_key2:my_value2"
+        self.assertRaisesRegexp(CloudifyCliError,
+                                expected_err_msg.format(input_str),
+                                utils.plain_string_to_dict,
+                                input_str)
+
+    def test_inputs_to_dict(self):
```
dankilman: perhaps rename this test method to be more clear. 
It doesn't actually test `inputs_to_dict` it tests `inputs_to_dict` error handling on an invalid key=value type string input

dankilman:
Maybe we should update all the help strings here
https://github.com/cloudify-cosmo/cloudify-cli/blob/master/cloudify_cli/config/parser_config.py
To remove the yaml mention.

PR #112, opened 2015-01-07
CFY-1910 added retry to container run execution and validation that does not exist prior to bootstapping.

adamlavie:
added retry to container run execution since the services container sporadically fails to mount the data container volumes. This also required adding a validation so that the bootstrap process would not kill containers that were not started by it.

### DIFF
--- a/cloudify_cli/bootstrap/tasks.py
+++ b/cloudify_cli/bootstrap/tasks.py
@@ -519,23 +519,29 @@ def _container_exists(docker_exec_command, container_name):
 
 
 def _run_docker_container(docker_exec_command, container_options,
-                          container_name, retries=1):
+                          container_name, retries_on_corrupt=1):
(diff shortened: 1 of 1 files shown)


cloudify_cli/bootstrap/tasks.py
```diff
@@ -498,6 +506,41 @@ def _run_command(command):
...
+    for i in range(0, retries):
+        try:
+            lgr.debug('starting docker container {0}'.format(container_name))
+            return _run_command(run_cmd)
+        except FabricTaskError:
+            lgr.debug('container execution failed on attempt {0}/{1}'
+                      .format(i + 1, retries))
+            if _container_exists(docker_exec_command, container_name):
+                rm_container_cmd = '{0} rm -f {1}'.format(docker_exec_command,
+                                                          container_name)
```
adamlavie: add else

PR #119, opened 2015-02-01
CFY-1747 added the missing modules to the local workflow ignore modules ...

yaelpeled:
...list.

### DIFF
--- a/cloudify_cli/tests/resources/blueprints/local/windows_installers_blueprint.yaml
+++ b/cloudify_cli/tests/resources/blueprints/local/windows_installers_blueprint.yaml
@@ -3,9 +3,13 @@ tosca_definitions_version: cloudify_dsl_1_0
 imports:
   - https://raw.githubusercontent.com/cloudify-cosmo/cloudify-manager/master/resources/rest-service/cloudify/types/types.yaml
 
+
 node_types:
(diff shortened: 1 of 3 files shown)


cloudify_cli/tests/resources/blueprints/local/windows_installers_blueprint.yaml
```diff
@@ -0,0 +1,30 @@
+tosca_definitions_version: cloudify_dsl_1_0
+
+imports:
+  - https://raw.githubusercontent.com/cloudify-cosmo/cloudify-manager/master/resources/rest-service/cloudify/types/types.yaml
```
ran-z: non-standard import; allegedly an import from S3 should be used. however in this case you probably want to avoid this altogether - you have no need for the actual base types, all you need for your test is an operation mapping to a task under the windows agent and windows plugin installers

if you still wish to use types.yaml, use the correct import + have install_agent: false on the host node for completeness of the test as an example

dankilman: I agree with Ran. Generally, tests that don't actually need types.yaml, better not import it. This is mainly because these are unit tests after all and they should only reach S3 (outside world) if absolutely necessary.

PR #124, opened 2015-02-23
Cfy 1838 split container support

nir0s:


barakm:
@adamlavie @nir0s what is the status here?

barakm:
@nir0s @adamlavie what is the status here?

nir0s:
This is irrelevant. We can close it.

PR #129, opened 2015-03-09
Cfy 1896 install agent with local workflow

yaelpeled:


cloudify_cli/tests/commands/test_local.py
```diff
@@ -20,6 +20,7 @@
 import os
 import json
 import tempfile
+from dsl_parser.parser import HOST_TYPE
```
dankilman: import should be places below

PR #132, opened 2015-03-12
CFY-2337: enable 'continue download functionality' in wget/curl

0lvin:
Use "continue download partial file feature" in wget/curl for minimize network traffic in case reattempt after failure operation.


nir0s:
Thank you for the pull request.. but the flow you're providing for is deprecated and will be removed with in 3.2.

Thank you.

PR #135, opened 2015-03-19
CFY-2353 use get_rest_client() instead of CloudifyClient

yaelpeled:


dankilman:
Fix already made in different commit

PR #149, opened 2015-04-05
Cfy 2251 add security framework token authentication

noakup:


cloudify_cli/utils.py
```diff
@@ -21,6 +21,8 @@
 import tempfile
 import getpass
 from contextlib import contextmanager
+from itsdangerous import base64_encode
```
noakup: should be in the 2nd bulk

PR #166, opened 2015-05-03
Cfy 1480 teardown after failure

iliapolo:


cosmo-admin:
@idanmo what is the status here? Is this still relevant?

PR #170, opened 2015-05-29
Standard version argument definition.

I159:
Issue
-----

Python SystemExit raises on every system exit event regardless of exit
code even a code is 0. This means exit message writes to stderr. A `Version`
action class uses parser.exit which does system exit, thus we have
version string written into stderr.

Solution
--------

To write a version string directly into stdout before system exit
occurred.

Issue 2
-------

If "--version" `arparse` argument defined in nonstandard way such as
custom actions it will break arguments parsing and cause known "too few
arguments" issue, when "-h" is becoming required argument.
http://stackoverflow.com/questions/19150074/argparse-optional-subparsers-with-store-true-flags

Solution to the issue 2
-----------------------

Use "--version" in standard way using "version" action and custom
function as `version` attribute. Considering the first issue the custom
function mustn't return a string but should write a version string
directly to stdout.
https://docs.python.org/3/library/argparse.html#action

Fixes Bug: https://cloudifysource.atlassian.net/browse/CFY-2773


barakm:
Not relevant anymore.

PR #183, opened 2015-07-14
CFY-3074 freeze docker versions

m1keil:


cloudify_cli/bootstrap/tasks.py
```diff
@@ -149,32 +150,27 @@ def _install_docker_if_required(docker_path, use_sudo,
             lgr.error(err)
             raise NonRecoverableError(err)
         try:
-            if current_distro == RHEL7X:
-                # install docker on RHEL 7.x.
-                # https://github.com/docker/docker/issues/11910
-                _run_command('{0} yum-config-manager --enable '
-                             'rhui-REGION-rhel-server-extras'.format(sudo))
-                _run_command('{0} yum install -y docker'.format(sudo))
+            if current_distro in [RHEL7X, CENTOS7X]:
```
nir0s: this is good.. but on a completely unrelated note, tasks.py has this function:

```python
def get_machine_distro():
    distro_info = _run_command('python -c "import platform, json, sys; '
                               'sys.stdout.write(\'{0}\\n\''
                               '.format(json.dumps(platform.dist())))"')
    return json.loads(distro_info)
```

I would replace `platform.dist()` with:
`platform.linux_distribution(full_distribution_name=False)` just because we know to avoid this problem.

cloudify_cli/bootstrap/tasks.py
```diff
@@ -149,32 +150,27 @@ def _install_docker_if_required(docker_path, use_sudo,
...
+                if current_distro == RHEL7X:
+                    # install docker on RHEL 7.x.
+                    # https://github.com/docker/docker/issues/11910
+                    _run_command('{0} yum-config-manager --enable '
+                                 'rhui-REGION-rhel-server-extras'.format(sudo))
+                _run_command('{0} yum install -y docker-1.6.2'.format(sudo))
```
nir0s: just wondering, is there no functional difference between 1.6.2 and 1.5 relevant for us?

cloudify_cli/bootstrap/tasks.py
m1keil: agree but i afraid to touch it

cloudify_cli/bootstrap/tasks.py
m1keil: not one that are forward incompatible

PR #184, opened 2015-07-15
CFY-3160 Test environment for plugin

geokala:


barakm:
@geokala please add some more details about what this command does

PR #198, opened 2015-08-10
Purge packaging binaries

nir0s:


barakm:
@nir0s what is the status here?

nir0s:
Needs to be redone.

PR #211, opened 2015-09-30
CFY-3036 Implement CLI commands for the snapshots API

xwombat:


nir0s:
please fix the tests. thanks.

cloudify_cli/commands/snapshots.py
```diff
@@ -0,0 +1,87 @@
...
+"""
+Handles all commands that start with 'cfy snapshots'
+"""
+
+from cloudify_cli import utils
+from cloudify_cli.logger import get_logger
+from cloudify_cli.utils import print_table
+
+
+def restore(snapshot_id, without_deployments_envs):
```
nir0s: @idanmo or @ran-z, in which scenario do we not want to restore deployment envs?

cloudify_cli/commands/snapshots.py
```diff
@@ -0,0 +1,87 @@
...
+    logger = get_logger()
```
nir0s: @ran-z, I now realize that we're doing the same across all CLI commands, but why are we generating the logger in each function? Why not set a globally accessible logger and, if needed, override it?

cloudify_cli/commands/snapshots.py
```diff
@@ -0,0 +1,87 @@
...
+    management_ip = utils.get_management_server_ip()
+    logger.info("Restoring snapshot '{0}' at management server {1}"
+                .format(snapshot_id, management_ip))
+    client = utils.get_rest_client(management_ip)
+    execution = client.snapshots.restore(snapshot_id,
```
nir0s: Um.. in create, we allow including metrics and excluding creds. It's rather strange that in restore we're not allowing to choose "items" up to the same granularity level. I might want to create snapshots with the metrics constantly but never restore them because while they're mission critical as archived info, I don't want them in my system when restoring. What I'm asking is.. aren't create and restore supposed to be symmetrical?

cloudify_cli/commands/snapshots.py
```diff
@@ -0,0 +1,87 @@
...
+                .format(snapshot_id, management_ip))
+    client = utils.get_rest_client(management_ip)
+    client.snapshots.delete(snapshot_id)
+    logger.info('Deleted snapshot successfully')
+
+
+def upload(snapshot_path, snapshot_id):
+    logger = get_logger()
+    management_ip = utils.get_management_server_ip()
+    logger.info("Uploading snapshot '{0}' at management server {1}"
```
nir0s: at -> to

cloudify_cli/commands/snapshots.py
```diff
@@ -0,0 +1,87 @@
...
+                .format(snapshot_path.name, management_ip))
+    client = utils.get_rest_client(management_ip)
+    snapshot = client.snapshots.upload(snapshot_path.name, snapshot_id)
+    logger.info("Uploaded snapshot, snapshot's id is: {0}"
```
nir0s: Uploaded snapshot with id: {0}

cloudify_cli/commands/snapshots.py
```diff
@@ -0,0 +1,87 @@
...
+    client = utils.get_rest_client(management_ip)
+    execution = client.snapshots.restore(snapshot_id,
+                                         not(without_deployments_envs))
+    logger.info('Started workflow\'s execution id: {0}'.format(execution.id))
+
+
+def create(snapshot_id, include_metrics, exclude_credentials):
+    logger = get_logger()
+    management_ip = utils.get_management_server_ip()
+    logger.info("Creating snapshot '{0}' at management server {1}"
```
nir0s: at -> in

cloudify_cli/commands/snapshots.py
```diff
@@ -0,0 +1,87 @@
...
+    client = utils.get_rest_client(management_ip)
+    snapshot = client.snapshots.upload(snapshot_path.name, snapshot_id)
+    logger.info("Uploaded snapshot, snapshot's id is: {0}"
+                .format(snapshot.id))
+
+
+def download(snapshot_id, output):
+    logger = get_logger()
+    management_ip = utils.get_management_server_ip()
+    logger.info("Downloading snapshot '{0}' ...".format(snapshot_id))
```
nir0s: This should probably be consistent with any other command and printout the management ip.

cloudify_cli/commands/snapshots.py
```diff
@@ -0,0 +1,87 @@
...
+    target_file = client.snapshots.download(snapshot_id, output)
+    logger.info("Snapshot '{0}' has been downloaded successfully as '{1}'"
+                .format(snapshot_id, target_file))
+
+
+def ls():
+    logger = get_logger()
+    management_ip = utils.get_management_server_ip()
+    client = utils.get_rest_client(management_ip)
+    logger.info('Getting snapshots list... [manager={0}]'
```
nir0s: Getting -> Retrieving

cloudify_cli/commands/snapshots.py
ran-z: @nir0s check out logger.py - "get_logger()" doesn't regenerate the logger, it's merely an abstraction for retrieving the logger which would otherwise in fact indeed be globally accessible.

cloudify_cli/config/parser_config.py
```diff
@@ -37,12 +37,24 @@ def blueprint_id_argument():
     }
 
 
+def snapshot_id_argument(hlp = 'The id of the snapshot'):
```
nir0s: can we remove this default value? It is generated elsewhere. Also, it isn't consistent with the rest of the argument definitions.

cloudify_cli/config/parser_config.py
```diff
@@ -37,12 +37,24 @@ def blueprint_id_argument():
...
+    return {
+        'metavar': 'SNAPSHOT_ID',
+        'type': str,
+        'help': hlp,
+        'dest': 'snapshot_id',
+        'default': None,
```
nir0s: @idanmo, maybe we want to allow users to, by default, act upon the latest snapshot?

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                            '-s,--snapshot-id': argument_utils.remove_completer(
+                                snapshot_id_argument(
+                                    hlp='A unique id that will be assigned to the created snapshot'
+                                )
+                            ),
+                            '--include-metrics': {
+                                'dest': 'include_metrics',
+                                'action': 'store_true',
+                                'default': False,
+                                'help': 'Specify this flag to include metrics data'
```
nir0s: This isn't directly related to this, but for this version, we should probably provide very clear documentation that users cannot backup large sets of metrics (like, 6 months worth or so). In the future, we should probably allows users to specify how many days they want to backup with a default of X days.

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                            },
+                            '--exclude-credentials': {
+                                'dest': 'exclude_credentials',
+                                'action': 'store_true',
+                                'default': False,
+                                'help': 'Specify this flag to prevent storing credentials'
+                                        'in the snapshot'
+                            }
+                        },
+                        'help': 'command for creating a new snapshots',
```
nir0s: "Creates a new snapshot."

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                                'action': 'store_true',
+                                'default': False,
+                                'help': 'Specify this flag to include metrics data'
+                                        'in the snapshot'
+                            },
+                            '--exclude-credentials': {
+                                'dest': 'exclude_credentials',
+                                'action': 'store_true',
+                                'default': False,
+                                'help': 'Specify this flag to prevent storing credentials'
```
nir0s: "Do not store credentials in the snapshot."

cloudify_cli/config/parser_config.py
nir0s: "Exclude metrics from snapshot."

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                            }
+                        },
+                        'help': 'command for creating a new snapshots',
+                        'handler': cfy.snapshots.create
+                    },
+                    'upload': {
+                        'arguments': {
+                            '-p,--snapshot-path': {
+                                'metavar': 'SNAPSHOT_FILE',
+                                'dest': 'snapshot_path',
```
nir0s: do we intend to let users provide a URL instead of a local path for uploading and restoring snapshots?

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                                'type': argparse.FileType(),
+                                'required': True,
+                                'help': "Path to the manager's snapshot file",
+                                'completer': completion_utils.yaml_files_completer
+                            },
+                            '-s,--snapshot-id': argument_utils.remove_completer(snapshot_id_argument())
+                        },
+                        'help': 'command for uploading a snapshot to the management server',
```
nir0s: "Upload a snapshot to the management server."

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                            '-s,--snapshot-id': snapshot_id_argument(),
+                            '-o,--output': {
+                                'metavar': 'OUTPUT',
+                                'type': str,
+                                'help': 'The output file path of the snapshot to be downloaded',
+                                'dest': 'output',
+                                'required': False
+                            }
+                        },
+                        'help': 'command for downloading a snapshot from the management server',
```
nir0s: "Download a snapshot from the management server."

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                        'handler': cfy.snapshots.download
+                    },
+                    'list': {
+                        'help': 'command for listing all snapshots on the '
```
nir0s: "List all snapshots on the management server."

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                                'Manager',
+                        'handler': cfy.snapshots.ls
+                    },
+                    'delete': {
+                        'arguments': {
+                            '-s,--snapshot-id': snapshot_id_argument()
+                        },
+                        'help': 'command for deleting a snapshot',
```
nir0s: "Delete a snapshot."

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                        'handler': cfy.snapshots.delete
+                    },
+                    'restore': {
+                        'arguments': {
+                            '-s,--snapshot-id': snapshot_id_argument(),
+                            '--without-deployments-envs': {
+                                'dest': 'without_deployments_envs',
+                                'action': 'store_true',
+                                'default': False,
+                                'help': 'Specify this flag to restore snapshot '
```
nir0s: "Restore snapshot without deployment environments."

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                                        'without deployments environments'
+                            }
+                        },
+                        'help': 'command for restoring manager to specific snapshot',
```
nir0s: "Restore a management server to a specific snapshot."

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                    }
+                }
+            },
+            'agents': {
+                'help': "Manages Cloudify's Agents",
+                'sub_commands': {
+                    'install': {
+                        'arguments': {
+                            '-d,--deployment-id': deployment_id_argument(
+                                hlp='The id of the deployment to install agents for, if not provided '
```
nir0s: ... agents for. If ommited, this will install..

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                                'then command will install agents for all deployments'
+                            ),
+                            '-l,--include-logs': {
+                                'dest': 'include_logs',
+                                'action': 'store_true',
+                                'help': 'A flag whether to include logs in returned events'
```
nir0s: "Whether to include logs in returned events or not."

cloudify_cli/config/parser_config.py
```diff
@@ -166,6 +178,111 @@ def parser_config():
...
+                            }
+                        },
+                        'help':'command for installing agents on deployments',
```
nir0s: "Install deployment agents for deployments."

cloudify_cli/commands/agents.py
```diff
@@ -0,0 +1,137 @@
...
+
+def install(deployment_id, include_logs):
+    workflow_id = 'install_new_agents'
+    logger = get_logger()
+    client = utils.get_rest_client()
+
+    if deployment_id:
+        deps = [deployment_id]
+        if not _deployment_exists(client, deployment_id):
+            logger.info("Deployment '{0}' does not exist"
```
nir0s: this should probably be logger.error (?)

cloudify_cli/commands/agents.py
```diff
@@ -0,0 +1,137 @@
...
+                        .format(deployment_id))
+            raise SuppressedCloudifyCliError()
+        if not _is_deployment_installed(client, deployment_id):
+            logger.info("Deployment '{0}' is not installed"
```
nir0s: same.

cloudify_cli/commands/agents.py
```diff
@@ -0,0 +1,137 @@
...
+                        .format(deployment_id))
+            raise SuppressedCloudifyCliError()
+        logger.info("Installing agent for deployment '{0}'"
+                    .format(deployment_id))
+    else:
+        deps = [dep.id for dep in client.deployments.list()
+                if _is_deployment_installed(client, dep.id)]
+        if not deps:
+            logger.info('There are no deployments installed')
```
nir0s: same.

cloudify_cli/commands/agents.py
```diff
@@ -0,0 +1,137 @@
...
+    threads = [threading.Thread(target=worker, args=(dep_id,))
+               for dep_id in deps]
+
+    for t in threads:
+        t.start()
+    for t in threads:
+        t.join()
+
+    if error_summary:
+        logger.info('Summary:\n{0}\n'.format(
```
nir0s: same.

cloudify_cli/commands/agents.py
```diff
@@ -0,0 +1,137 @@
...
+    for node_instance in client.node_instances.list(deployment_id):
+        if node_instance.state != _NODE_INSTANCE_STATE_STARTED:
+            return False
+    return True
+
+
+def _deployment_exists(client, deployment_id):
+    try:
+        client.deployments.get(deployment_id)
+    except:
```
nir0s: are we sure that we don't need to be more explicit here? The fact that we didn't manage to .get(deployment_id) necessarily means that it doesn't exist?

cloudify_cli/commands/agents.py
nir0s: if we do need to be more explicit but there's no way for us to currently identify whether we failed to get the deployments or that they simply don't exist, maybe the message should say: `"Could not retrieve deployment for deployment_id: {0}.".format(deployment_id)`

cloudify_cli/commands/agents.py
```diff
@@ -0,0 +1,137 @@
...
+from cloudify_cli.exceptions import SuppressedCloudifyCliError
+from cloudify_cli.exceptions import ExecutionTimeoutError
+from cloudify import logs
+
+_NODE_INSTANCE_STATE_STARTED = 'started'
+
+
+def _is_deployment_installed(client, deployment_id):
+    for node_instance in client.node_instances.list(deployment_id):
+        if node_instance.state != _NODE_INSTANCE_STATE_STARTED:
```
nir0s: @idanmo Just making sure that we know that this will return False if we're in the middle of an install workflow. Since this practically checks for an arbitrary node's state.

cloudify_cli/tests/commands/test_snapshots.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def test_snapshots_restore(self):
+        self.client.snapshots.restore = MagicMock()
+        cli_runner.run_cli('cfy snapshots restore -s a-snapshot-id')
+        cli_runner.run_cli('cfy snapshots restore -s a-snapshot-id'
+                           '--without-deployments-workers')
+
+    def test_snapshots_download(self):
+        self.client.snapshots.download = MagicMock(return_value='some_file')
+        cli_runner.run_cli('cfy snapshots download -s a-snapshot-id')
```
nir0s: ehh... there are no tests for errors here like when a deployment_id doesn't exist or a snapshot file is missing... or.. otherwise.

cloudify_cli/config/parser_config.py
konradxyz: If I understand correctly "host agents" then it is exactly what it does.

It installs new agents on application vms and makes sure those agents are connected to current manager.

It is a part of agent upgrade process.

nir0s: It's my bad. ignore.

cloudify_cli/commands/snapshots.py
nir0s: my bad.

cloudify_cli/commands/snapshots.py
xwombat: We were asked to add flags only to create command. If symmetry is something that you want then we can add it in different pull request.

cloudify_cli/config/parser_config.py
xwombat: For this one and all comments below that want to change subcommand help message:
we use help format 'command for ...' because it is consistent with other help messages for subcommands, like in blueprints and deployments. I don't think we want break that.

cloudify_cli/config/parser_config.py
xwombat: Implementing snapshots I modeled on blueprints implementation. In blueprints you are allowed to provide URL so I add that option also to snapshots.

cloudify_cli/commands/agents.py
xwombat: Agents install command should be used directly after snapshots restore. Even if user create/install new deployments in the meantime, then install command should be used only on deployments that were restored from snapshot. Any other scenario is considered as error.

cloudify_cli/config/parser_config.py
xwombat: What do you mean 'Exclude metrics from snapshots.". Metrics are excluded by default.

cloudify_cli/commands/snapshots.py
konradxyz: It was useful for development purposes. It is a lot of faster to restore snapshot without deployment env installation.

simon-bar: Additionally, sometimes you might just want to "tidy up" your ES data when it gets corrupt, without touching dep envs (it lasts long and apart from that you must delete those envs before).

cloudify_cli/config/parser_config.py
nir0s: Just.. to replace the "help" message with this.

cloudify_cli/commands/snapshots.py
nir0s: thanks.

cloudify_cli/commands/snapshots.py
nir0s: Alright. Thank you.

cloudify_cli/config/parser_config.py
nir0s: Alright. Thank you

cloudify_cli/config/parser_config.py
xwombat: Ok, but flag says that metrics will be included. So maybe help message should be "Include metrics to snapshot"?

PR #231, opened 2015-11-12
CFY-4067 - Added flag --force to 'snapshots restore'.

xwombat:


cloudify_cli/config/parser_config.py
```diff
@@ -328,6 +328,13 @@ def parser_config():
                                 'action': 'store_true',
                                 'default': False,
                                 'help': 'Restore snapshot without deployment environments'
+                            },
+                            '--force': {
+                                'dest': 'force',
+                                'action': 'store_true',
+                                'default': False,
+                                'help': 'Specify this flag to skip checking '
```
idanmo: Please change help string to: "Force restoring the snapshot on a dirty manager"

PR #248, opened 2015-12-16
CFY-4066 Remove client side error handling for event errors

dankilman:


cloudify_cli/execution_events_fetcher.py
```diff
@@ -43,19 +42,12 @@ def _fetch_and_process_events_batch(self, events_handler=None):
...
-            events, total = self._client.events.get(
-                self._execution_id,
-                from_event=self._from_event,
-                batch_size=self._batch_size,
-                include_logs=self._include_logs)
-            self._from_event += len(events)
-        except CloudifyClientError, e:
-            # A workaround in a case where events index was not yet created.
-            # This can happen if there were no events sent to logstash.
-            if e.status_code == 500 and 'IndexMissingException' in e.message:
```
ran-z: is this really the same error? i dont see "IndexMissingException" in the message in the CFY-4066 jira
just want to make sure we're not removing a workaround for a different error..

PR #272, opened 2016-01-26
CFY-4393 Delay fetching events after an execution ended

pawelcodilime:


cloudify_cli/execution_events_fetcher.py
```diff
@@ -19,6 +19,7 @@
 
 
 WAIT_FOR_EXECUTION_SLEEP_INTERVAL = 3
+WAIT_BEFORE_FETCHING_LAST_EVENTS_SLEEP_INTERVAL = 2
```
max-orlov: I assume this entire pr is no longer relevant...

tehasdf:
this is obsoleted by #303 which resoled CFY-4393 in a different way

PR #275, opened 2016-02-08
CFY-4503 - added --validate for blueprints upload, default is to skip validations

isaac-s:


ran-z:
should add simple unit tests which checks usage of the flag
(a test which passes the flag and an invalid blueprint and "succeeds", another one which doesnt pass the flag and fails, or vice versa depends on the default behavior etc.)

isaac-s:
@ran-z added test cases

PR #294, opened 2016-03-17
CFY-5051 CFY 5051 show default argument values in the CLI

AviaE:


AviaE:
Reviewed by @dankilman

PR #300, opened 2016-03-24
CFY-1599 cfy executions list/get add message if status is canceled

liormizr:


cloudify_cli/commands/executions.py
```diff
@@ -29,6 +29,11 @@
 from cloudify_cli.logger import get_events_logger
 
 
+_STATUS_CANCELING_MESSAGE = (
+    'NOTE: Executions currently in a "canceling status" '
+    'may take a while to complete')
+
```
liormizr: NOTE: Executions currently in a "canceling/force-canceling" status may take a while to change into "cancelled"

PR #302, opened 2016-03-28
CFY-5101 fix interactive shell

nir0s:


cloudify_cli/commands/ssh.py
```diff
@@ -28,6 +30,17 @@
 from cloudify_cli.exceptions import CloudifyCliError
 
 
+def _open_interactive_shell(host_string, command=''):
+    """Used as fabric's open_shell=True doesn't work well.
+    """
+    cmd = ['ssh', '-t']
+    cmd.append(host_string)
+    cmd.extend(['-i', os.path.expanduser(utils.get_management_key())])
+    if command:
```
liormizr: create the cmd list like this:
cmd = [
    'ssh', '-t', host_string,
    '-i', os.path.expanduser(utils.get_management_key()),
]
You want to do the most when you create the list,
You want to modify the list as little as possible...

cloudify_cli/commands/ssh.py
```diff
@@ -28,6 +30,17 @@
 from cloudify_cli.exceptions import CloudifyCliError
 
 
+def _open_interactive_shell(host_string, command=''):
```
liormizr: I would have changed the command arg to tmux=False
so the api is: _open_interactive_shell(host_string, tmux=False)
and the tmux command defined inside the method...

cloudify_cli/commands/ssh.py
```diff
@@ -20,6 +20,8 @@
 import platform
 from distutils import spawn
 import re
+import subprocess
+import os
```
liormizr: I like to sort the imports by length, but it's not critical...
import os
import re
import platform
...
from distutils import spawn

PR #303, opened 2016-03-29
CFY-4393 wait for a finished event when running an execution

tehasdf:
Running an execution like `cfy executions start` only waits until
the execution's status changes to terminated (or failed, cancelled).

This makes it so the cli also waits for a workflow_succeeded (or failed,
cancelled) event, not just for the status change.

### DIFF
--- a/cloudify_cli/execution_events_fetcher.py
+++ b/cloudify_cli/execution_events_fetcher.py
@@ -19,6 +19,8 @@ from cloudify_rest_client.executions import Execution
 
 
 WAIT_FOR_EXECUTION_SLEEP_INTERVAL = 3
+WORKFLOW_END_TYPES = {u'workflow_succeeded', u'workflow_failed',
+                      u'workflow_cancelled'}
 
 
 class ExecutionEventsFetcher(object):
@@ -85,6 +87,34 @@ def get_deployment_environment_creation_execution(client, deployment_id):
                            executions))
 
 
+class EventsWatcher(object):
+    """Wraps an event_handler function, examines events to check if an
+    workflow execution finished has arrived.
+
+    This will set its .end_log_received instance attribute to True,
+    when it receives an event of type workflow_succeeded, workflow_cancelled
+    or workflow_failed.
+
+    :ivar end_log_received: was a "workflow execution finished" event seen?
+    :vartype end_log_received: bool
+    """
+
+    def __init__(self, events_handler=None):
+        self._events_handler = events_handler
+        self.end_log_received = False
+
+    def __call__(self, events):
+        if self._events_handler is not None:
+            self._events_handler(events)
+
+        if any(self._is_end_event(evt) for evt in events):
+            self.end_log_received = True
+
+    def _is_end_event(self, event):
+        """Is event a 'workflow execution finished' event?"""
+        return event.get('event_type') in WORKFLOW_END_TYPES
+
+
 def wait_for_execution(client,
                        execution,
                        events_handler=None,
@@ -101,7 +131,10 @@ def wait_for_execution(client,
     events_fetcher = ExecutionEventsFetcher(client, execution.id,
                                             include_logs=include_logs)
 
-    # Poll for execution status until execution ends
+    # Poll for execution status and execution logs, until execution ends
+    # and we receive an event of type in WORKFLOW_END_TYPES
+    execution_ended = False
+    events_watcher = EventsWatcher(events_handler)
     while True:
         if timeout is not None:
             if time.time() > deadline:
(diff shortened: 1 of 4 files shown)


cloudify_cli/execution_events_fetcher.py
```diff
@@ -101,7 +131,10 @@ def wait_for_execution(client,
     events_fetcher = ExecutionEventsFetcher(client, execution.id,
                                             include_logs=include_logs)
 
-    # Poll for execution status until execution ends
+    # Poll for execution status and execution logs, until execution ends
+    # and we receive an event of type in WORKFLOW_END_TYPES
+    execution_ended = False
+    events_watcher = EventsWatcher(events_handler)
     while True:
```
max-orlov: How do you feel placing "not(execution_ended and events_watcher.end_log_received)" as the condition instead of just True (since this block obviously has to end)?

cloudify_cli/execution_events_fetcher.py
```diff
@@ -112,18 +145,21 @@ def wait_for_execution(client,
...
                 # update the remaining timeout
-                timeout = deadline-time.time()
+                timeout = deadline - time.time()
 
-        if execution.status != Execution.PENDING:
-            events_fetcher.fetch_and_process_events(
-                events_handler=events_handler, timeout=timeout)
+        if not execution_ended:
+            execution = client.executions.get(execution.id)
+            if execution.status in Execution.END_STATES:
```
max-orlov: I think execution_ended = execution.status in Execution.END_STATES could simplify things, what do you think?

tehasdf: yep that's right, thanks :D

cloudify_cli/execution_events_fetcher.py
tehasdf: hmm, not sure how to write that, while still keeping the original behaviour, and making it more readable

all changes would change the behaviour of the function slightly, maybe only in unrealistic hypothetical cases, but still...

a few possible variants in pseudocode here https://gist.github.com/tehasdf/c65a833321ed4e15566eec75e871ce89
i could rewrite it as the while_condition_sleep_first version version but i'm not sure if it's really worth it

tehasdf:
Updated the PR, fixing 2 tests that were broken by this change

cloudify_cli/execution_events_fetcher.py
max-orlov: Yeah, you'r right. I've added a comment to the gist, though again, you are right, it might not be as cleaner as I've hoped originally...

tehasdf:
Added a new commit (will squash later if needed), rewriting event tests to not use threads (they were sometimes failing on CI, even though they passed on my machine...); this should make them easier to understand and debug, and less brittle; please review.

Although CI is still failing right now due to broken dev-requirements.txt on master :(

cloudify_cli/tests/commands/test_events.py
```diff
@@ -31,34 +30,35 @@ def mock_log_message_prefix(event):
...
-        events_generator = threading.Thread(
-            target=self.generate_events,
-            args=(self.execution_termination_time,))
-        events_generator.daemon = True
-        events_generator.start()
-
-    def generate_events(self, execution_termination_time):
-        while time.time() < execution_termination_time:
-            # to simulate a common events flow, sleep for 3 secs every 2 secs
-            if len(EventsTest.events) > 0 and time.time() % 2 == 0:
```
tehasdf: btw this didn't even really work: `time.time()` is a float and the chance of that %2 to equal exactly 0.0 is.. low :D

cloudify_cli/tests/commands/test_events.py
```diff
@@ -79,10 +80,8 @@ def _mock_executions_get(self, execution_id):
 
     def _mock_events_get(self, execution_id, from_event=0,
                          batch_size=100, include_logs=False):
-        if from_event >= len(self.events):
-            return [], len(self.events)
-        until_event = min(from_event + batch_size, len(self.events))
-        return self.events[from_event:until_event], len(self.events)
+        events = self._get_events_before(time.time())
+        return events[from_event:], len(events)
```
max-orlov: what happend to the batch_size? previously it was used to measure the 'until_event', it seems it's missing now

tehasdf: Good catch! Fixed now. (the `min` is redundant since slicing beyond the bounds would just return only up to the end of the list)

Otherwise I think the function should be understandable? In the previous/threaded version, one thread was generating the events, and another was running the function under test that polled for events. This version pre-generates all the events, and only returns those that come with a timestamp smaller than "now".

Btw the function under test is still doing a `sleep()`, as it was before.
We could also mock out the `time.sleep` and `time.time` calls so that then the tests would run instantly, rather than in 12 seconds :)
But I didn't want to add another behaviour change to this PR

max-orlov: yeah, good call, looks good =] squash and merge :) ?

PR #306, opened 2016-04-13
CFY-5071 Version compatibility between Cloudify Manager and CLI tool …

pawelcodilime:
…is not policed

### DIFF
--- a/cloudify_cli/utils.py
+++ b/cloudify_cli/utils.py
@@ -18,6 +18,7 @@ import glob
 import json
 import os
 import pkgutil
+import socket
 import sys
 import tempfile
 import getpass
@@ -29,6 +30,7 @@ from contextlib import contextmanager
 
 import yaml
 import pkg_resources
+from cloudify_rest_client.exceptions import CloudifyClientError
 from jinja2.environment import Template
 from prettytable import PrettyTable
 from itsdangerous import base64_encode
@@ -324,7 +326,8 @@ def get_cwd():
     return os.getcwd()
 
 
-def get_rest_client(manager_ip=None, rest_port=None, protocol=None):
+def get_rest_client(manager_ip=None, rest_port=None, protocol=None,
+                    skip_version_check=False):
     if not manager_ip:
         manager_ip = get_management_server_ip()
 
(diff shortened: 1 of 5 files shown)


cloudify_cli/utils.py
```diff
@@ -344,8 +347,21 @@ def get_rest_client(manager_ip=None, rest_port=None, protocol=None):
...
+
+    cli_version, manager_version = get_cli_manager_versions()
+
+    if cli_version == manager_version:
+        return client
+    else:
+        message = ('CLI and manager versions do not match\n'
+                   'Cloudify CLI: {}\n'
+                   'Cloudify Manager: {}').format(cli_version, manager_version)
+        sys.exit(message)
```
nir0s: Is there a reason for sys.exit here instead of a Cloudify CLI Exception?

cloudify_cli/utils.py
```diff
@@ -344,8 +347,21 @@ def get_rest_client(manager_ip=None, rest_port=None, protocol=None):
...
+
+    if skip_version_check:
+        return client
+
+    cli_version, manager_version = get_cli_manager_versions()
+
+    if cli_version == manager_version:
+        return client
+    else:
+        message = ('CLI and manager versions do not match\n'
```
nir0s: capital M please.

cloudify_cli/utils.py
```diff
@@ -344,8 +347,21 @@ def get_rest_client(manager_ip=None, rest_port=None, protocol=None):
...
+                   'Cloudify CLI: {}\n'
```
nir0s: ...CLI version:
...Manager version:

cloudify_cli/utils.py
```diff
@@ -344,8 +347,21 @@ def get_rest_client(manager_ip=None, rest_port=None, protocol=None):
...
+                   'Cloudify Manager: {}').format(cli_version, manager_version)
```
nir0s: Right now, our convention is to keep compatibility with python 2.6 in any new code. While the CLI itself doesn't fully support 2.6, we should still keep that compat until decided otherwise. {} will fail on string formatting. please use {0}

cloudify_cli/utils.py
pawelcodilime: There isn't any particular reason. Should I use CloudifyCliError (I assume that you meant this exception)?

nir0s:
LGTM, did you run a test to verify that this works?

PR #309, opened 2016-04-25
CFY-4880 CLI message about "execution timed out" is misleading

pawelcodilime:


cloudify_cli/commands/agents.py
```diff
@@ -116,8 +116,10 @@ def worker(dep_id):
                                " '{1}'".format(workflow_id, dep_id))
 
         except ExecutionTimeoutError as e:
-            log_to_summary("Execution of workflow '{0}' "
+            log_to_summary("Waiting for execution of workflow '{0}' "
```
isaac-s: I suggest the following text (for `agents.py` and `executions.py`):

Timed out waiting for workflow '{0}' of deployment '{1}' to end. The execution may still be running properly; however, the command-line utility was instructed to wait up to {3} seconds for its completion.

* Run 'cfy executions list' to determine the execution's status.
* Run 'cfy executions cancel --execution-id {2}' to cancel the running workflow.

PR #317, opened 2016-05-15
CFY-4954 added cli command

adamlavie:
CFY-4954 add cli manager upgrade command

### DIFF
--- a/cloudify_cli/commands/upgrade.py
+++ b/cloudify_cli/commands/upgrade.py
@@ -1,0 +1,118 @@
+########
+# Copyright (c) 2016 GigaSpaces Technologies Ltd. All rights reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+#    * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    * See the License for the specific language governing permissions and
+#    * limitations under the License.
+
+"""
+Handles 'cfy upgrade command'
+"""
+import time
+import json
+import tempfile
+
+from cloudify_cli import ssh
+from cloudify_cli import utils
+from cloudify_cli import common
+from cloudify_cli import exceptions
+from cloudify_cli.logger import get_logger
+from cloudify_cli.commands import maintenance
+
+
+MAINTENANCE_MODE_DEACTIVATED = 'deactivated'
+MAINTENANCE_MODE_ACTIVATING = 'activating'
+
+REMOTE_WORKFLOW_STATE_PATH = '/opt/manager/_workflow_state.json'
+
+
+def upgrade(validate_only,
+            skip_validations,
+            blueprint_path,
+            inputs,
+            install_plugins,
+            task_retries,
+            task_retry_interval,
+            task_thread_pool_size):
+
+    logger = get_logger()
+    management_ip = utils.get_management_server_ip()
+
+    client = utils.get_rest_client(management_ip)
+    curr_status = client.maintenance_mode.status().status
+    if curr_status == MAINTENANCE_MODE_DEACTIVATED:
+        msg = "'Manager must be in maintenance mode for upgrade to run."
+        error = exceptions.CloudifyCliError(msg)
+        error.possible_solutions = [
+            "Activate maintenance mode by running "
+            "'cfy maintenance-mode activate'"
+        ]
+        raise error
+    elif curr_status == MAINTENANCE_MODE_ACTIVATING:
+        _wait_for_maintenance(client, logger)
+
+    if not inputs:
+        inputs = {}
+    inputs_data = utils.inputs_to_dict(inputs, 'inputs') or {}
+    if not inputs_data.get('private_ip'):
+        raise exceptions.CloudifyCliError('Private IP must be provided for '
+                                          'the upgrade process')
+    if 'ssh_key_filename' not in inputs_data.keys():
+        inputs_data.update({'ssh_key_filename': utils.get_management_key()})
+    if 'public_ip' not in inputs_data.keys():
+        inputs_data.update({'public_ip': management_ip})
+    if 'ssh_user' not in inputs_data.keys():
+        inputs_data.update({'ssh_user': utils.get_management_user()})
+
+    env_name = 'manager-upgrade'
+    # init local workflow execution environment
+    env = common.initialize_blueprint(blueprint_path,
+                                      storage=None,
+                                      install_plugins=install_plugins,
+                                      name=env_name,
+                                      inputs=str(inputs_data))
+
+    if not skip_validations:
+        logger.info('Executing upgrade validations...')
+        env.execute(workflow='execute_operation',
+                    parameters={'operation':
+                                'cloudify.interfaces.validation.creation'},
+                    task_retries=task_retries,
+                    task_retry_interval=task_retry_interval,
+                    task_thread_pool_size=task_thread_pool_size)
+
+    if not validate_only:
+        logger.info('Starting manager upgrade process...')
+        _put_workflow_state_file(is_upgrade=True)
+
+        logger.info('Executing manager upgrade...')
+        env.execute('install',
+                    task_retries=task_retries,
+                    task_retry_interval=task_retry_interval)
+
+    logger.info('Deactivating manager maintenance-mode...')
+    client.maintenance_mode.deactivate()
+
+
+def _put_workflow_state_file(is_upgrade):
+    manager_state_file = tempfile.NamedTemporaryFile(delete=True)
+    content = {'is_upgrade': is_upgrade}
+    with open(manager_state_file.name, 'w') as f:
+        f.write(json.dumps(content))
+    ssh.put_file_in_manager(manager_state_file, REMOTE_WORKFLOW_STATE_PATH)
+
+
+def _wait_for_maintenance(client, logger):
+    curr_status = client.maintenance_mode.status().status
+    while curr_status != maintenance.MAINTENANCE_MODE_ACTIVE:
+        logger.info('Waiting for maintenance mode activation...')
+        time.sleep(maintenance.DEFAULT_TIMEOUT_INTERVAL)
+        curr_status = client.maintenance_mode.status().status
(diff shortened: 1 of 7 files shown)


cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+                        task_retry_interval=task_retry_interval)
+        except Exception as e:
+            logger.info('Upgrade process failed. Starting manager '
+                        'rollback')
+            raise e
+
+    logger.info('Deactivating manager maintenance-mode')
+    client.maintenance_mode.deactivate()
+
+
```
adamlavie: This will replace the current way of detecting whether we are in upgrade/rollback state.

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+#    * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    * See the License for the specific language governing permissions and
+#    * limitations under the License.
+
+"""
+Handles 'cfy upgrade command'
+"""
+import yaml
```
nir0s: imports should be separated:

```python
stdlib

external

cloudify

same package
```

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+
+    management_ip = utils.get_management_server_ip()
+    client = utils.get_rest_client(management_ip)
+
+    curr_status = client.maintenance_mode.status().status
+    if curr_status == MAINTENANCE_MODE_DEACTIVATED:
+        msg = "'Manager must be in maintenance mode for upgrade to run."
+        error = exceptions.CloudifyCliError(msg)
+        error.possible_solutions = [
+            "Activate maintenance mode using 'cfy maintenance-mode activate'"
```
nir0s: using -> by running

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+    env_name = 'manager-upgrade'
+    # init local workflow execution environment
+    env = common.initialize_blueprint(blueprint_path,
+                                      storage=None,
+                                      install_plugins=install_plugins,
+                                      name=env_name,
+                                      inputs=inputs)
+
+    if not skip_validations:
+        logger.info('Executing upgrade validation')
```
nir0s: `Executing upgrade validations...`

tox.ini
```diff
@@ -8,7 +8,7 @@ deps =
 
 [testenv:py27]
 deps =
-    coverage==3.7.1 # this fixes issue with tox installing coverage --pre
```
nir0s: ?

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+            logger.info('Executing manager upgrade.')
+            env.execute('install',
+                        task_retries=task_retries,
+                        task_retry_interval=task_retry_interval)
+        except Exception as e:
+            logger.info('Upgrade process failed. Starting manager '
+                        'rollback')
+            raise e
+
+    logger.info('Deactivating manager maintenance-mode')
```
nir0s: mode...

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+        logger.info('Starting manager upgrade process.')
+        _put_workflow_state_file(fabric_data, is_upgrade=True)
+        try:
+            logger.info('Executing manager upgrade.')
+            env.execute('install',
+                        task_retries=task_retries,
+                        task_retry_interval=task_retry_interval)
+        except Exception as e:
+            logger.info('Upgrade process failed. Starting manager '
+                        'rollback')
```
nir0s: rollback...

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+            raise e
```
nir0s: do we also want to expose the exception in the log message?

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+        with open(inputs) as f:
+            inputs_data = yaml.load(f)
+        fabric_data = {'host_string': management_ip,
+                       'user': inputs_data['ssh_user'],
+                       'key_filename': inputs_data['ssh_key_filename']}
+
+        logger.info('Starting manager upgrade process.')
+        _put_workflow_state_file(fabric_data, is_upgrade=True)
+        try:
+            logger.info('Executing manager upgrade.')
```
nir0s: upgrade...

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+                    task_thread_pool_size=task_thread_pool_size)
+
+    if not validate_only:
+        with open(inputs) as f:
+            inputs_data = yaml.load(f)
+        fabric_data = {'host_string': management_ip,
+                       'user': inputs_data['ssh_user'],
+                       'key_filename': inputs_data['ssh_key_filename']}
+
+        logger.info('Starting manager upgrade process.')
```
nir0s: process...

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+        logger.info('Executing upgrade validation')
+        env.execute(workflow='execute_operation',
+                    parameters={'operation':
+                                'cloudify.interfaces.validation.creation'},
+                    task_retries=task_retries,
+                    task_retry_interval=task_retry_interval,
+                    task_thread_pool_size=task_thread_pool_size)
+
+    if not validate_only:
+        with open(inputs) as f:
```
nir0s: so, two things. one, inputs are not always a file. they can be multiple files, json, key value, etc. to handle these inputs, there is a util for that:

```python
inputs = utils.inputs_to_dict(inputs, 'inputs')
```

on top of that, I don't quite understand why you need inputs at all. While we might want to allow overriding all inputs, the `ssh_user` and `ssh_key_filename` are exposed using:

```python
ssh_key_path = utils.get_management_key()
username = utils.get_management_user()
```

This will save the user from having to provide those inputs when upgrading. Instead, they could be default just run `cfy upgrade -p ...`

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+
+import fabric.api
+import fabric.context_managers
+
+MAINTENANCE_MODE_DEACTIVATED = 'deactivated'
+MAINTENANCE_MODE_ACTIVATING = 'activating'
+
+REMOTE_WORKFLOW_STATE_PATH = '/opt/manager/_workflow_state.json'
+
+logger = get_logger()
```
nir0s: you can't initialize the logger globally. It will override all other loggers.

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def _put_workflow_state_file(fabric_data, is_upgrade):
+    manager_state_file = tempfile.NamedTemporaryFile(delete=True)
+    content = {'is_upgrade': is_upgrade}
+    with open(manager_state_file.name, 'w') as f:
+        f.write(json.dumps(content))
+
+    with _manager_fabric_env(**fabric_data) as api:
+        api.put(use_sudo=True,
```
nir0s: `ssh.py` contains a function called `get_file_from_manager`. Please add another function there called `put_file_in_manager` which implements the same logic so that you don't have to implement it specifically here.
Since it already includes the logic for retrieving everything like host_string, etc.. you won't have to use the fabric data thing above.

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+                local_path=manager_state_file,
+                remote_path=REMOTE_WORKFLOW_STATE_PATH)
+
+
+def _wait_for_maintenance(client):
+    curr_status = client.maintenance_mode.status().status
+    while curr_status != maintenance.MAINTENANCE_MODE_ACTIVE:
+        logger.info('Waiting for maintenance mode activation.')
```
nir0s: activation...

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+        time.sleep(maintenance.DEFAULT_TIMEOUT_INTERVAL)
+        curr_status = client.maintenance_mode.status().status
+
+
+@contextmanager
```
nir0s: this can be ditched if implementing the above

cloudify_cli/commands/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+def _manager_fabric_env(**kwargs):
```
nir0s: this can be ditched if implementing the above

cloudify_cli/config/parser_config.py
```diff
@@ -1067,6 +1066,36 @@ def parser_config():
                 },
                 'handler': cfy.bootstrap
             },
+            'upgrade': {
+                'help': 'Upgrade manager ',
+                'arguments': {
+                    '-p,--blueprint-path':
+                        local_blueprint_path_argument(
+                                hlp='Path to a blueprint'
```
nir0s: It's not JUST a blueprint...

cloudify_cli/config/parser_config.py
```diff
@@ -1067,6 +1066,36 @@ def parser_config():
...
+                        ),
+                    '-i,--inputs': {'dest': 'inputs',
+                                            'help': '',
+                                            },
+                    '--skip-validations': {
+                        'dest': 'skip_validations',
+                        'action': 'store_true',
+                        'help': 'Run bootstrap without '
```
nir0s: not bootstrap

cloudify_cli/config/parser_config.py
```diff
@@ -1067,6 +1066,36 @@ def parser_config():
...
+                                'validating resources prior to bootstrapping the manager'
```
nir0s: not bootstrap

cloudify_cli/config/parser_config.py
```diff
@@ -1067,6 +1066,36 @@ def parser_config():
...
+                    },
+                    '--validate-only': {
+                        'dest': 'validate_only',
+                        'action': 'store_true',
+                        'help': 'Run validations without '
+                                'actually performing the bootstrap process'
```
nir0s: not bootstrap

tox.ini
adamlavie: Tox upgraded their version and inline comments are forbidden

cloudify_cli/commands/upgrade.py
adamlavie: This should not be here. removing try/catch..

PR #319, opened 2016-05-18
CFY-5285 Handle SIGINT gracefully in `cfy agents install`

tehasdf:
.join() on a thread uses a lock that cannot be interrupted by
signals. Instead, use the "while True: poll(); sleep()" pattern
to allow the cli to exit on user ctrl+c

### DIFF
--- a/cloudify_cli/commands/agents.py
+++ b/cloudify_cli/commands/agents.py
@@ -17,11 +17,13 @@
 Handles all commands that start with 'cfy agents'
 """
 
+import time
 import threading
 
 from cloudify_cli import utils
 from cloudify_cli.logger import get_logger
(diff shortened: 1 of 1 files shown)


cloudify_cli/commands/agents.py
```diff
@@ -133,9 +135,13 @@ def worker(dep_id):
                for dep_id in deps]
 
     for t in threads:
+        t.daemon = True
         t.start()
-    for t in threads:
-        t.join()
+
+    while True:
```
tehasdf: No timeout handling is needed here, because all the worker threads already have a timeout - so the whole operation will time out when all the worker thread executions do.

PR #320, opened 2016-05-18
CFY-5326 refine and standardize CLI help texts

nir0s:


cloudify_cli/config/parser_config.py
```diff
@@ -34,7 +34,9 @@
 from cloudify_cli.constants import DEFAULT_INSTALL_WORKFLOW
 from cloudify_cli.constants import DEFAULT_UNINSTALL_WORKFLOW
 
-FORMAT_INPUT_AS_YAML_OR_DICT = 'formatted as YAML or as "key1=value1;key2=value2"'
+FORMAT_INPUT_AS_YAML_OR_DICT = \
+    ('Can be provided as wildcard based paths to YAML files, '
```
AviaE: What are these `wildcard based paths`? What is their format, and how the user is supposed to know about it?

cloudify_cli/config/parser_config.py
```diff
@@ -86,15 +88,15 @@ def validate_blueprint_argument():
         'dest': 'validate_blueprint',
         'action': 'store_true',
         'help': 'Validate the blueprint before uploading it '
-                'to the manager'
+                'to Cloudify Manager'
```
AviaE: This is the only time that you refer to the Manager as 'Cloudify Manager' and not as 'Manager'. Why?

cloudify_cli/config/parser_config.py
```diff
@@ -168,7 +170,7 @@ def timeout_argument(default_timeout=DEFAULT_TIMEOUT):
         'type': int,
         'default': default_timeout,
         'help': 'Operation timeout in seconds (The execution itself will keep '
-                'going. It is the CLI that will stop waiting for it to '
+                'going but the CLI will stop waiting for it to '
```
AviaE: Maybe a comma after 'going'?

cloudify_cli/config/parser_config.py
nir0s: good point.

cloudify_cli/config/parser_config.py
```diff
@@ -244,26 +246,27 @@ def auto_generate_ids_argument():
 
 def parser_config():
     return {
-        'description': 'Manages Cloudify in different Cloud Environments',
+        'description': 'A Command Line Interface Cloudify',
```
AviaE: This change is great in concept. 
But what about something in the likes of "Cloudify's Command Line Interface"? I'm not familiar with the sentence structuring you proposed.

cloudify_cli/config/parser_config.py
nir0s: yup

cloudify_cli/config/parser_config.py
nir0s: wildcard based paths are obvious to anyone using bash... but I'll add an example in parentheses

cloudify_cli/config/parser_config.py
nir0s: it was supposed to be '...for Cloudify' .. but I'm fine with what you're proposing

cloudify_cli/config/parser_config.py
```diff
@@ -276,22 +279,22 @@ def parser_config():
...
                                 'dest': 'backup_first',
-                                'help': 'Whether to backup before purging.'
-                                        'Backup will be in tar.gz format.',
+                                'help': 'Whether to backup before purging. '
+                                        'Backup will be in tar.gz format',
                                 'action': 'store_true',
                             }
                         },
-                        'help': "Delete a Manager's logs",
+                        'help': "Delete the Manager's service logs",
```
AviaE: Is there a difference between the Manager's 'logs' and its 'service logs'? Why does the 'purge' command specifically refers to 'service logs', while the two other sub commands under logs refer to 'logs'?

cloudify_cli/config/parser_config.py
```diff
@@ -312,18 +315,17 @@ def parser_config():
...
-                                        ' creation ({0}). '
+                        inputs_argument('Inputs for the deployment ({0}). '
                                         'This argument can be used multiple times. '
                                         '(default: {1})'
                                         .format(FORMAT_INPUT_AS_YAML_OR_DICT,
                                                 DEFAULT_INPUTS_PATH_FOR_INSTALL_COMMAND)
                                         ),
                     '-w,--workflow': make_optional(workflow_id_argument(
-                            hlp='The workflow to start '
+                            hlp='A workflow name to execute '
```
AviaE: It sound a bit like we are executing the workflow's name, but I can't think of a better suggestion.

cloudify_cli/config/parser_config.py
```diff
@@ -339,17 +341,16 @@ def parser_config():
...
                     '-d,--deployment-id': make_required(
                             deployment_id_argument(
-                                hlp='The id of the deployment you wish to '
+                                hlp='The ID of the deployment you wish to '
                                     'uninstall'
                             )
                         ),
                     '-w,--workflow': make_optional(workflow_id_argument(
-                            hlp='The workflow to start (default: {0}'
+                            hlp='A workflow name to execute (default: {0}'
```
AviaE: Same as above

cloudify_cli/config/parser_config.py
nir0s: It should probably be "The name of the workflow to execute"

cloudify_cli/config/parser_config.py
nir0s: because it was 1:30am when I did this :)

cloudify_cli/config/parser_config.py
```diff
@@ -400,24 +401,24 @@ def parser_config():
...
                     'list': {
-                        'help': 'List all the plugins on the Manager',
+                        'help': 'List all plugins currently on the Manager',
                         'handler': cfy.plugins.ls
                     },
                     'delete': {
                         'arguments': {
                             '-p,--plugin-id': plugin_id_argument(
-                                hlp='The plugin id'),
+                                hlp="A plugin's ID"),
```
AviaE: Why 'A' and not 'The'?
Notice that in 'cfy snapshots delete', "--snapshot-id"'s help text is:
'The ID of the snapshot to delete'
I suggest that you decide on a format right now, and then go ahead and apply it all over this module =)

cloudify_cli/config/parser_config.py
```diff
@@ -448,7 +449,7 @@ def parser_config():
                         'arguments': {
                             '-b,--blueprint-id': blueprint_id_argument(),
                             '-o,--output': {
-                                'help': 'The output file path of the blueprint to be downloaded',
+                                'help': 'The local path of the downloaded blueprint',
```
AviaE: `cfy plugins download`'s `--output` argument states:
'A local path for the downloaded plugin',
and here `cfy blueprints download`'s `--output` states:
'**The** local path **of** the downloaded blueprint

cloudify_cli/config/parser_config.py
```diff
@@ -244,26 +247,27 @@ def auto_generate_ids_argument():
...
+                'help': 'show version information and exit',
                 'action': cfy.version
             }
         },
         'commands': {
             'logs': {
-                'help': 'Handle Cloudify Manager logs',
+                'help': "Handle the Manager's logs",
                 'sub_commands': {
                     'get': {
```
AviaE: Compare `cfy blueprints download --output` and `cfy plugins download --output`, which, according to their help texts, seem to serve about the same purpose. So why here it's `logs --destination-path`?

cloudify_cli/config/parser_config.py
nir0s: I thought I did. That's why I asked you to check for inconsistencies :) It was late at night. I'll do that right now. `The` actually makes more sense.

cloudify_cli/config/parser_config.py
```diff
@@ -522,7 +522,7 @@ def parser_config():
                                 'dest': 'snapshot_path',
                                 'type': argparse.FileType(),
                                 'required': True,
-                                'help': "Path to the manager's snapshot file",
+                                'help': "Local path to the Manager's snapshot file",
```
AviaE: Add `A`\`The` as usual

cloudify_cli/config/parser_config.py
```diff
@@ -548,75 +548,74 @@ def parser_config():
...
                             },
                             '-f,--force':
                                 force_argument(
                                         hlp='Force restoring the snapshot on '
                                             'a Manager with existing '
-                                            'blueprints and/or deployments')
+                                            'blueprints or deployments')
                         },
-                        'help': 'Restore manager state to a specific snapshot',
+                        'help': "Restore a Manager using a snapshot",
```
AviaE: Why did you change it to double quotes?

cloudify_cli/config/parser_config.py
```diff
@@ -548,75 +548,74 @@ def parser_config():
...
                         'handler': cfy.snapshots.restore
                     }
                 }
             },
             'agents': {
-                'help': "Manage Cloudify's Agents",
+                'help': "Handle a deployment's agents",
```
AviaE: Great. Much more informative.

cloudify_cli/config/parser_config.py
```diff
@@ -548,75 +548,74 @@ def parser_config():
...
+                'help': "Handle deployments on the Manager",
                 'sub_commands': {
                     'create': {
                         'arguments': {
                             '-d,--deployment-id': make_required(
                                 remove_completer(
                                     deployment_id_argument(
-                                        hlp='A unique id that will be assigned'
-                                            ' to the created deployment'
+                                        hlp='A unique ID to assign to the created deployment'
```
AviaE: compare with `cfy snapshots create --snapshot-id`: 
`A unique ID for the snapshot`
BTW, isn't an ID unique by definition?

cloudify_cli/config/parser_config.py
nir0s: let's forget the implementation for a second and concentrate on the help texts. If it bothers you, please open a JIRA issue for it.

cloudify_cli/config/parser_config.py
nir0s: good q

cloudify_cli/config/parser_config.py
```diff
@@ -668,43 +666,42 @@ def parser_config():
...
                             '--system-workflows': {
                                 'dest': 'include_system_workflows',
                                 'action': 'store_true',
-                                'help': 'Include executions of '
-                                        'system workflows'
+                                'help': 'Include executions of system workflows'
                             },
                         },
-                        'help': 'List all running executions, or all '
+                        'help': 'List all running executions on the Manager or all '
```
AviaE: Does `cfy execution list` list all **running** executions, or all executions, including those who already terminated?
If only running executions, then maybe change `or all executions for a specific deployment` to `or all **running** executions for a specific deployment`
If not only running executions, then delete the `running`.

cloudify_cli/config/parser_config.py
```diff
@@ -718,42 +715,41 @@ def parser_config():
...
                         },
-                        'help': 'Cancel an execution by its id',
+                        'help': 'Cancel an execution',
                         'handler': cfy.executions.cancel
                     }
                 }
             },
             'nodes': {
-                'help': 'Manage nodes',
+                'help': "Handle a blueprint's nodes",
```
AviaE: `deployment's nodes` instead of `blueprint's nodes`?

cloudify_cli/config/parser_config.py
```diff
@@ -773,61 +769,61 @@ def parser_config():
...
                         'arguments': {
                             '--node-instance-id': {
                                 'dest': 'node_instance_id',
                                 'required': True,
-                                'help': 'The ID of the node instance to get'
+                                'help': 'The ID of the node-instance to get'
                             }
                         },
-                        'help': "Get a node instance according to its ID",
+                        'help': "Get a node instance",
```
AviaE: add hyphen to `node instance`

cloudify_cli/config/parser_config.py
nir0s: oops

cloudify_cli/config/parser_config.py
```diff
@@ -1032,28 +1024,27 @@ def parser_config():
...
-                        'help': 'If the bootstrap fails,'
-                                ' the Manager will remain running'
+                        'help': 'Do not teardown the Manager even if the bootstrap fails'
                     },
                     '--skip-validations': {
                         'dest': 'skip_validations',
                         'action': 'store_true',
-                        'help': 'Run bootstrap without '
-                                'validating resources prior to bootstrapping the manager'
+                        'help': 'Run a bootstrap without '
```
AviaE: Maybe without the `Run a` part. Just `Bootstrap without validating ...`.

cloudify_cli/config/parser_config.py
```diff
@@ -1100,25 +1090,24 @@ def parser_config():
...
             'teardown': {
-                'help': 'Teardown Cloudify',
+                'help': 'Teardown the Manager',
                 'arguments': {
                     '--ignore-deployments': {
                         'dest': 'ignore_deployments',
                         'action': 'store_true',
-                        'help': 'Perform teardown even if deployments'
-                                'exist on the manager'
+                        'help': 'Perform the teardown even if there are '
```
AviaE: Just like the previous comment, maybe start the help text with `Teardown`

PR #321, opened 2016-05-19
CFY-5306 rollback cli command

adamlavie:


cloudify_cli/bootstrap/bootstrap.py
```diff
@@ -514,6 +514,13 @@ def upload_plugin(plugin_path):
                             plugins.validate)
 
     for plugin_source_path in plugin_resources:
+        logger = get_logger()
+        plugin_exists = rest_client.plugins.list(
+                archive_name=os.path.basename(plugin_source_path)).items
```
adamlavie: Get the actual archive name from the plugin metadata file

cloudify_cli/commands/rollback.py
```diff
@@ -0,0 +1,62 @@
...
+
+    env_name = 'manager-rollback'
+    # init local workflow execution environment
+    env = common.initialize_blueprint(blueprint_path,
+                                      storage=None,
+                                      install_plugins=install_plugins,
+                                      name=env_name,
+                                      inputs=json.dumps(inputs))
+
+    logger.info('Starting manager rollback process...')
```
adamlavie: Does the rollback workflow need to execute implicitly?

cloudify_cli/commands/rollback.py
```diff
@@ -0,0 +1,62 @@
...
+             install_plugins,
+             task_retries,
+             task_retry_interval):
+
+    logger = get_logger()
+    management_ip = utils.get_management_server_ip()
+
+    client = utils.get_rest_client(management_ip, skip_version_check=True)
+
+    handle_maintenance_mode_activation(client, logger)
```
adamlavie: use get_logger from within

cloudify_cli/commands/upgrade.py
```diff
@@ -46,68 +48,124 @@ def upgrade(validate_only,
...
-    if not inputs_data.get('private_ip'):
-        raise exceptions.CloudifyCliError('Private IP must be provided for '
-                                          'the upgrade process')
-    if 'ssh_key_filename' not in inputs_data.keys():
-        inputs_data.update({'ssh_key_filename': utils.get_management_key()})
-    if 'public_ip' not in inputs_data.keys():
-        inputs_data.update({'public_ip': management_ip})
-    if 'ssh_user' not in inputs_data.keys():
-        inputs_data.update({'ssh_user': utils.get_management_user()})
+    handle_maintenance_mode_activation(client, logger)
```
adamlavie: verify_and_waitfor

cloudify_cli/commands/upgrade.py
```diff
@@ -46,68 +48,124 @@ def upgrade(validate_only,
...
+                            user=inputs['ssh_user'])
     if not skip_validations:
         logger.info('Executing upgrade validations...')
-        env.execute(workflow='execute_operation',
-                    parameters={'operation':
-                                'cloudify.interfaces.validation.creation'},
-                    task_retries=task_retries,
-                    task_retry_interval=task_retry_interval,
-                    task_thread_pool_size=task_thread_pool_size)
+        try:
```
adamlavie: remove the try catch

cloudify_cli/commands/upgrade.py
```diff
@@ -46,68 +48,124 @@ def upgrade(validate_only,
...
+    inputs.update({'ssh_key_filename': _load_management_key(inputs)})
+    inputs.update({'ssh_user': _load_management_user(inputs)})
+    inputs.update({'public_ip': utils.get_management_server_ip()})
+    return inputs
+
+
+def _load_private_ip(inputs):
+    try:
+        if inputs.get('private_ip'):
+            return inputs.get('private_ip')
```
adamlavie: fix

adamlavie: how about removing the sec get?

cloudify_cli/commands/upgrade.py
```diff
@@ -46,68 +48,124 @@ def upgrade(validate_only,
...
-                    task_thread_pool_size=task_thread_pool_size)
+        try:
+            env.execute(workflow='execute_operation',
+                        parameters={'operation':
+                                    'cloudify.interfaces.validation.creation'},
+                        task_retries=task_retries,
+                        task_retry_interval=task_retry_interval,
+                        task_thread_pool_size=task_thread_pool_size)
+            logger.info('Upgrade validation completed successfully')
+        except Exception as e:
```
adamlavie: move to  the install workflow exec

cloudify_cli/config/parser_config.py
```diff
@@ -1065,11 +1065,12 @@ def parser_config():
                 'arguments': {
```
adamlavie: no space at the end of help

cloudify_cli/config/parser_config.py
```diff
@@ -1089,6 +1090,23 @@ def parser_config():
                 },
                 'handler': cfy.upgrade.upgrade
             },
+            'rollback': {
+                'help': 'rollback manager upgrade',
```
adamlavie: prettify

cloudify_cli/ssh.py
```diff
@@ -23,11 +23,16 @@ def get_file_from_manager(remote_source_path, destination_path):
...
 
-def put_file_in_manager(source_path, remote_source_path, use_sudo=True):
-    key_filename = os.path.expanduser(utils.get_management_key())
+def put_file_in_manager(source_path,
+                        remote_source_path,
+                        use_sudo=True,
+                        key_filename=None,
+                        user=None):
+    if not key_filename:
+        key_filename = os.path.expanduser(utils.get_management_key())
```
adamlavie: expand user anyways

cloudify_cli/bootstrap/bootstrap.py
```diff
@@ -514,6 +514,13 @@ def upload_plugin(plugin_path):
...
+        if plugin_exists:
+            logger.info('Plugin with archive name {0} already exists on the '
```
nir0s: "The plugin {0} already exists on the Manager. Skipping upload..."

cloudify_cli/commands/rollback.py
nir0s: Manager

cloudify_cli/commands/rollback.py
```diff
@@ -0,0 +1,62 @@
...
+    # init local workflow execution environment
+    env = common.initialize_blueprint(blueprint_path,
+                                      storage=None,
+                                      install_plugins=install_plugins,
+                                      name=env_name,
+                                      inputs=json.dumps(inputs))
+
+    logger.info('Starting manager rollback process...')
+    put_workflow_state_file(is_upgrade=False,
+                            key_filename=inputs['ssh_key_filename'],
```
nir0s: again, these can be taken from the local context instead of inputs.. as they must be the same anyway.

cloudify_cli/commands/rollback.py
```diff
@@ -0,0 +1,62 @@
...
+                            user=inputs['ssh_user'])
+
+    logger.info('Executing manager rollback...')
```
nir0s: Manager

cloudify_cli/commands/rollback.py
```diff
@@ -0,0 +1,62 @@
...
+    env.execute('install',
```
nir0s: There is no obvious precedence for doing this, but it would be safer if we assumed that there could be an error here while upgrading.. in which case...

```python
try:
    env.execute(...)
    logger.info('Rollback complete...')
except Exception as ex:
    logger.error('...')
```

cloudify_cli/commands/upgrade.py
```diff
@@ -46,68 +48,124 @@ def upgrade(validate_only,
...
                     task_retries=task_retries,
-                    task_retry_interval=task_retry_interval)
+                    task_retry_interval=task_retry_interval,
+                    task_thread_pool_size=task_thread_pool_size)
 
-    logger.info('Deactivating manager maintenance-mode...')
-    client.maintenance_mode.deactivate()
+        fabric_env = {
+            "host_string": utils.get_management_server_ip(),
+            "user": inputs['ssh_user'],
```
nir0s: same as above

cloudify_cli/config/parser_config.py
nir0s: Please make sure all help texts you add here are perfectly consistent with other help texts in the CLI.

cloudify_cli/commands/rollback.py
adamlavie: They are taken from the local context. Checkout 'update_inputs' function

cloudify_cli/commands/upgrade.py
adamlavie: Read 'update_inputs'

cloudify_cli/commands/rollback.py
adamlavie: why?

cloudify_cli/commands/rollback.py
adamlavie: ?

cloudify_cli/utils.py
```diff
@@ -678,3 +679,12 @@ def build_manager_host_string(user='', ip=''):
     user = user or get_management_user()
     ip = ip or get_management_server_ip()
     return '{0}@{1}'.format(user, ip)
+
+
+def load_plugin_package_details(plugin_path):
+    with tarfile.open(plugin_path, 'r') as tar:
+        tar_members = tar.getmembers()
+        package_json_path = '{0}/package.json'.format(tar_members[0].name)
```
nir0s: this will break on windows

cloudify_cli/config/parser_config.py
```diff
@@ -1089,6 +1090,23 @@ def parser_config():
                 },
                 'handler': cfy.upgrade.upgrade
             },
+            'rollback': {
+                'help': 'Rollback the Manager upgrade',
```
nir0s: "Rollback the Manager to its previous version" ?

cloudify_cli/commands/upgrade.py
```diff
@@ -88,26 +73,103 @@ def upgrade(validate_only,
...
+    except Exception:
+        raise exceptions.CloudifyCliError('Manager user must be provided for '
+                                          'the upgrade/rollback process')
+
+
+def verify_and_wait_for_maintenance_mode_activation(client):
+    logger = get_logger()
+    curr_status = client.maintenance_mode.status().status
+    if curr_status == MAINTENANCE_MODE_DEACTIVATED:
+        msg = "Manager must be in maintenance mode for workflow to run."
```
nir0s: remove period for consistency

nir0s: maintenance-mode. Nir Biran is adding dashes everywhere

cloudify_cli/commands/upgrade.py
```diff
@@ -88,26 +73,103 @@ def upgrade(validate_only,
...
+    except Exception:
+        raise exceptions.CloudifyCliError('Management key must be provided for'
+                                          ' the upgrade/rollback process')
+
+
+def _load_management_user(inputs):
+    try:
+        if inputs.get('ssh_user'):
+            return inputs.get('ssh_user')
+        return utils.get_management_user()
```
nir0s:
```python
try:
    return inputs.get('ssh_user', utils.get_management_user())
except: ...
```

cloudify_cli/commands/upgrade.py
```diff
@@ -88,26 +73,103 @@ def upgrade(validate_only,
...
+                                          ' the upgrade/rollback process')
+
+
+def _load_management_key(inputs):
+    try:
+        if inputs.get('ssh_key_filename'):
+            key_path = inputs.get('ssh_key_filename')
+        else:
+            key_path = utils.get_management_key()
+        return os.path.expanduser(key_path)
```
nir0s: I assume ssh_key_filename can't be None or False or 0 or empty string. Is that right?
in which case..

```python
key_path = inputs.get('ssh_key_filename', utils.get_management_key())
```

cloudify_cli/commands/upgrade.py
```diff
@@ -88,26 +73,103 @@ def upgrade(validate_only,
...
+    inputs.update({'private_ip': _load_private_ip(inputs)})
+    inputs.update({'ssh_key_filename': _load_management_key(inputs)})
+    inputs.update({'ssh_user': _load_management_user(inputs)})
+    inputs.update({'public_ip': utils.get_management_server_ip()})
+    return inputs
+
+
+def _load_private_ip(inputs):
+    try:
+        if inputs.get('private_ip'):
```
nir0s: same as below

cloudify_cli/commands/upgrade.py
```diff
@@ -88,26 +73,103 @@ def upgrade(validate_only,
...
-    logger.info('Deactivating manager maintenance-mode...')
-    client.maintenance_mode.deactivate()
+        try:
+            logger.info('Executing Manager upgrade...')
+            env.execute('install',
+                        task_retries=task_retries,
+                        task_retry_interval=task_retry_interval,
+                        task_thread_pool_size=task_thread_pool_size)
+        except Exception as e:
+            msg = "'Failed upgrading Manager. Error: {0}".format(e.message)
```
nir0s: remove single quote

nir0s: use str(e)

cloudify_cli/bootstrap/bootstrap.py
```diff
@@ -514,6 +514,17 @@ def upload_plugin(plugin_path):
...
+            _get_resource_into_dir(temp_dir, plugin_source_path,
+                                   retries, wait_interval, timeout)
+        package = utils.load_plugin_package_details(plugin_local_path)
+        plugin_exists = rest_client.plugins.list(
+                archive_name=package['archive_name']).items
+        if plugin_exists:
+            logger.info('The plugin {0} already exists on the manager. '
+                        'Skipping upload...'.format(plugin_source_path))
+            continue
         plugin_local_path = \
```
dankilman: you can remove this as this duplicate

cloudify_cli/bootstrap/bootstrap.py
```diff
@@ -514,6 +514,17 @@ def upload_plugin(plugin_path):
                             plugins.validate)
 
     for plugin_source_path in plugin_resources:
+        logger = get_logger()
+        plugin_local_path = \
+            _get_resource_into_dir(temp_dir, plugin_source_path,
+                                   retries, wait_interval, timeout)
+        package = utils.load_plugin_package_details(plugin_local_path)
+        plugin_exists = rest_client.plugins.list(
```
dankilman: maybe make one rest call to read all plugins in the beginning to avoid making a rest call for each plugin resource and then filter on the client side?

cloudify_cli/commands/rollback.py
```diff
@@ -0,0 +1,70 @@
...
+                            user=inputs['ssh_user'])
+
+    logger.info('Executing Manager rollback...')
+    try:
+        env.execute('install',
+                    task_retries=task_retries,
+                    task_retry_interval=task_retry_interval)
+    except Exception as e:
+        msg = 'Failed to rollback Manager upgrade. Error: {0}'\
+            .format(e.message)
```
dankilman: `format(e)` instead of `format(e.message)`.

cloudify_cli/commands/upgrade.py
dankilman: don't use str(e), use e as it is already part of format so it gets called implicitly

cloudify_cli/commands/upgrade.py
```diff
@@ -88,26 +73,103 @@ def upgrade(validate_only,
...
+            error = exceptions.CloudifyCliError(msg)
+            error.possible_solutions = [
+                "Rerun Manager upgrade command 'cfy upgrade'",
+                "Execute rollback command 'cfy rollback'"
+            ]
+            raise error
+
+        fabric_env = {
+            "host_string": utils.get_management_server_ip(),
```
dankilman: `management_ip` is already a variable in the function, no need to call get_management_server_ip again

cloudify_cli/commands/upgrade.py
```diff
@@ -88,26 +73,103 @@ def upgrade(validate_only,
...
+        return load_env().outputs()['private_ip']
+    except Exception:
+        raise exceptions.CloudifyCliError('Private IP must be provided for'
+                                          ' the upgrade/rollback process')
+
+
+def _load_management_key(inputs):
+    try:
+        if inputs.get('ssh_key_filename'):
+            key_path = inputs.get('ssh_key_filename')
```
dankilman: no need for get inside the if block

cloudify_cli/utils.py
dankilman: that is incorrect. the path is inside the tar, not the file system

cloudify_cli/commands/upgrade.py
adamlavie: This is wrong. If we use it like you suggest, the 'utils.get_management_key()' will be evaluated before calling the get and will raise an exception.

cloudify_cli/commands/upgrade.py
adamlavie: Same as above. This will raise an exception

cloudify_cli/commands/upgrade.py
adamlavie: Same

cloudify_cli/config/parser_config.py
adamlavie: This may be a little misleading since it's an upgrade rollback. I do agree that there could be a better phrasing though...

cloudify_cli/commands/upgrade.py
tehasdf: you can avoid evaluating that if you use something like `inputs.get('ssh_user') or utils.get_management_user()` - this pattern is also used in other places (in `utils.build_manage_host_string`)

cloudify_cli/bootstrap/bootstrap.py
```diff
@@ -513,12 +513,29 @@ def upload_plugin(plugin_path):
...
     for plugin_source_path in plugin_resources:
         plugin_local_path = \
             _get_resource_into_dir(temp_dir, plugin_source_path,
                                    retries, wait_interval, timeout)
-
-        upload_plugin(plugin_local_path)
+        package = utils.load_plugin_package_details(plugin_local_path)
+        local_plugin_packages[plugin_local_path] = package['archive_name']
+
+    existing_plugins = []
```
dankilman: use `set` instead of `list`

cloudify_cli/commands/rollback.py
```diff
@@ -0,0 +1,69 @@
...
+                            user=inputs['ssh_user'])
+
+    logger.info('Executing Manager rollback...')
+    try:
+        env.execute('install',
+                    task_retries=task_retries,
+                    task_retry_interval=task_retry_interval)
+    except Exception as e:
+        msg = 'Failed to rollback Manager upgrade. Error: {0}'.format(e)
+        error = exceptions.CloudifyCliError(msg)
```
dankilman: bug here
simply do
`raise exceptions.CloudifyCliError(msg)`

cloudify_cli/commands/upgrade.py
```diff
@@ -88,26 +73,96 @@ def upgrade(validate_only,
...
+            raise error
+
+        fabric_env = {
+            "host_string": management_ip,
+            "user": inputs['ssh_user'],
+            "key_filename": inputs['ssh_key_filename']
+        }
+        manager_node = next(node for node in env.storage.get_nodes()
+                            if node.id == 'manager_configuration')
+        bs.upload_resources(manager_node, fabric_env, inputs['public_ip'],
```
tehasdf: Please note that this fires requests against the REST service, so perhaps maintenance mode should be deactivated first.

cloudify_cli/commands/upgrade.py
```diff
@@ -88,26 +72,96 @@ def upgrade(validate_only,
...
+            ]
+            raise error
+
+        manager_node = next(node for node in env.storage.get_nodes()
+                            if node.id == 'manager_configuration')
+        plugin_resources = \
+            manager_node.properties['cloudify'].get('upload_resources', {})\
+                        .get('plugin_resources', ())
+        if plugin_resources:
+            logger.info('Plugins upload is not supported for upgrade. Plugins '
```
dankilman: warn instead of info

dankilman: we should also log a warning if the user provided dsl_resources

PR #363, opened 2016-07-06
CFY-3691 - secure internal requests

noakup:


cloudify_cli/utils.py
```diff
@@ -560,19 +584,19 @@ class CloudifyWorkingDirectorySettings(yaml.YAMLObject):
     yaml_loader = yaml.Loader
 
     def __init__(self):
-        self._management_ip = None
```
dankilman: management_host

PR #368, opened 2016-07-14
raising uninitialized error if calling cfy use before init

noakup:


geokala:
Looks reasonable to me, I think (though I don't know this bit of the test framework very well, so I may have missed something).

PR #373, opened 2016-07-18
CFY-5509 Allow uploading plugins from a URL

mcouthon:
The infrastructure was already there - I just made sure the local validation wouldn't be triggered in case the plugin path is a URl


mcouthon:
@nir0s regarding tests, I see that the `cli_runner` isn't up and running, and all the tests seem to be using it, so I'm not sure whether I should write tests that would use it, or what.

mcouthon:
Commited directly to cfy-redo

PR #395, opened 2016-08-02
Cfy redo todos

mcouthon:


cloudify_cli/tests/commands/test_base.py
```diff
@@ -15,7 +15,8 @@
 import os
 import os as utils_os
 
-import testtools
+# import testtools
+import unittest  # TODO: replace after we're done testing
```
nir0s: Won't work. many env related tests require this

PR #398, opened 2016-08-02
Changed how the blueprint_path and name are being retrieved

mcouthon:


cloudify_cli/utils.py
```diff
@@ -147,6 +148,8 @@ def generate_suffixed_id(id):
 
 
 def is_archive(source):
+    if not source:
```
nir0s: Um.. in what situation will someone call this without pass a source?

mcouthon: I had a case where I could pass None to it (in install.py)

PR #402, opened 2016-08-04
Check the file before reading it

yuexiao-wang:
It should check the file whether it exist before reading it.
@idanmo PTAL, 
Thanks!


nir0s:
Thanks. A part of our refactoring in the `cfy-redo` branch takes care of many of these situations.

PR #416, opened 2016-08-10
CFY-5827 Allow customizing the install_agents.py script location

tehasdf:


cloudify_cli/commands/agents.py
```diff
@@ -95,9 +95,18 @@ def worker(dep_id):
         timeout = 900
 
         try:
+            kwargs = {}
+            if install_script is not None:
+                kwargs = {
+                    'parameters': {
+                        'install_script': install_script
+                    },
+                    'allow_custom_parameters': True
```
tehasdf: preexisting deployments don't have the `install_script` parameter in their workflow config, so it's considered a custom parameter for them

PR #435, opened 2016-08-30
CFY-5894 log tracebacks to file regardless of verbosity level

nir0s:


cloudify_cli/cli/cfy.py
```diff
@@ -150,11 +150,18 @@ def recommend(possible_solutions):
             logger.info('  - {0}'.format(solution))
 
     def new_excepthook(tpe, value, tb):
+        traceback.print_exception(
+            etype=tpe,
+            value=value,
+            tb=tb,
+            file=open(DEFAULT_LOG_FILE, 'a'))
```
ran-z: is this someething standard?
shouldnt it be something like
with open(....) as f:
  traceback.print_exception(.... file=f)?

PR #445, opened 2016-09-26
fix use function call in teardown

Jessiebr:


geokala:
Closed: As this PR is more than six months old without having been merged. Please reopen (and rebase if necessary) if it is still relevant.

PR #450, opened 2016-10-04
Cfy 6006. Use requests instead of urllib

jcollado:
In this PR, `requests` is used as suggested in the issue instead of `urllib` to see if it solves the problem found when downloading a blueprint file.


cloudify_cli/utils.py
```diff
@@ -158,20 +159,43 @@ def unzip(archive, destination=None):
...
     if not destination:
         fd, destination = tempfile.mkstemp()
         os.close(fd)
     logger = get_logger()
     logger.info('Downloading {0} to {1}...'.format(url, destination))
-    final_url = urllib.urlopen(url).geturl()
+
+    try:
+        response = requests.get(url, stream=True)
+    except requests.exceptions.RequestException as exception:
```
idanmo: For consistency please use "ex" instead of "exception"

cloudify_cli/utils.py
```diff
@@ -158,20 +159,43 @@ def unzip(archive, destination=None):
...
         logger.debug('Redirected to {0}'.format(final_url))
-    f = urllib.URLopener()
+
     try:
-        f.retrieve(final_url, destination)
-    except IOError as ex:
+        with open(destination, 'wb') as destination_file:
+            for chunk in response.iter_content(CHUNK_SIZE):
+                destination_file.write(chunk)
+    except IOError as exception:
```
idanmo: ex...

setup.py
```diff
@@ -49,6 +49,7 @@
         'jinja2==2.7.2',
         'retrying==1.3.3',
         'colorama==0.3.3',
+        'requests==2.7.0',
```
idanmo: Why did you choose specifically this version of requests?

cloudify_cli/utils.py
jcollado: Fixed in bc06628

cloudify_cli/utils.py
jcollado: Fixed in bc06628

setup.py
jcollado: It's the version that was already installed in my virtual environment because is required in other repositories like:  `cloudify-dsl-parser`, `cloudify-manager`, `cloudify-rest-client`, ...

In case we decide to go for the latest version, the requirement should probably be updated in all the related repositories to avoid requiring different versions in the same virtual environment.

jcollado:
@idanmo With regard to the squashing of the commits, I'll do that in the github UI once the PR is approved since that's more convenient (it ends up with just one commit, not with a merge commit wrapping the squashed commit).

PR #452, opened 2016-10-08
CFY-6006 Retry download without SSL verification on error

jcollado:
In this PR, when an attempt to download a file fails because of an SSL verification problem, a retry will be made with the verification step disabled. Note that this is just a workaround and a final fix will remove this behavior in the future.


jcollado:
The way to fix the problem will be to let the manager download the blueprint file instead of the client as it happened in the 3.4 version. Hence, this PR to workaround the download problem in the client is no longer relevant.

PR #453, opened 2016-10-09
CFY-5979 Add tenants commands

dusking:


cloudify_cli/cli/helptexts.py
```diff
@@ -162,3 +162,5 @@
 
 INSTALL_SCRIPT_LOCATION = \
     'Alternative location of the `install_agents.py` script'
+TENANT = 'The name of the tenant'
```
dusking: Maybe other text?

PR #457, opened 2016-10-19
CFY-6006 Use publish_archive instead of upload when a url is passed

jcollado:
When a URL is passed, the `client.blueprints.upload` call is no longer
useful because it expects a file to be passed. The right API call in
this case is `client.blueprints.publish_archive`.

### DIFF
--- a/cloudify_cli/commands/blueprints.py
+++ b/cloudify_cli/commands/blueprints.py
@@ -134,9 +134,9 @@ def upload(ctx,
                 progress_handler,
             )
         finally:
-            # Every situation other than the user providing a path of a local
-            # yaml means a temp folder will be created that should be later
-            # removed.
+            # When an archive file is passed, it's extracted to a temporary
+            # directory to get the blueprint file. Once the blueprint has been
+            # uploaded, the temporary directory needs to be cleaned up.
             if processed_blueprint_path != blueprint_path:
                 temp_directory = os.path.dirname(
                     os.path.dirname(processed_blueprint_path)
--- a/cloudify_cli/commands/blueprints.py
+++ b/cloudify_cli/commands/blueprints.py
@@ -18,6 +18,8 @@ import os
 import json
 import shutil
 
+from urlparse import urlparse
+
 import click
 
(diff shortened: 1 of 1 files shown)


cloudify_cli/commands/blueprints.py
```diff
@@ -96,28 +98,53 @@ def upload(ctx,
...
+
+    progress_handler = utils.generate_progress_handler(blueprint_path, '')
+    blueprint_id = blueprint_id or blueprint.generate_id(
+        processed_blueprint_path, blueprint_filename)
+
+    if is_url:
+        # When a URL is passed it's assumed to be pointing to an archive
+        # file that contains the blueprint. Hence, the `publish_archive`
+        # API call is the one that should be used.
+        logger.info('Publishing blueprint archive %s...', blueprint_path)
```
idanmo: Please be consistent with string formatting. {0} etc...

cloudify_cli/commands/blueprints.py
```diff
@@ -96,28 +98,53 @@ def upload(ctx,
...
+        try:
+            if validate:
+                ctx.invoke(
+                    validate_blueprint,
+                    blueprint_path=processed_blueprint_path,
+                )
+
+            # When the blueprint file is already available locally, it can be
+            # uploaded directly using the `upload` API call.
+            logger.info('Uploading blueprint %s...', blueprint_path)
```
idanmo: String format...

cloudify_cli/commands/blueprints.py
jcollado: I usually prefer using `.format`, but note that in this case the string interpolation doesn't happen in this line but in the logging code. This is considered to be a good practice in python because if the logging level is set to, let's say, `warning` the interpolation won't happen at all because the message isn't printed.

jcollado: For completeness, you can find a brief explanation of the problem in the [pylint documentation](https://pylint.readthedocs.io/en/latest/reference_guide/features.html#messages) (in the *logging-not-lazy (W1201)* bullet).

PR #458, opened 2016-10-19
CFY-6005. Fix use callback invocation

jcollado:
In this PR, the `use` command that was called directly from the `teardown` one is now called using `ctxt.invoke` to avoid parsing arguments from the command line.

Note: I had to include 2832295 in this branch make the test cases pass. However, the commit is not in the `master` branch. I believe the commit should be pushed to the `master` branch before this branch, so that the diff in this branch contains only the changes related to the issue that has been fixed.

### DIFF
--- a/cloudify_cli/tests/commands/test_teardown.py
+++ b/cloudify_cli/tests/commands/test_teardown.py
@@ -1,5 +1,3 @@
-import sys
-
 from mock import patch, MagicMock
 
 from .. import env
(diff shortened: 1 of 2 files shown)


jcollado:
@idanmo Now that the version update is in master, this PR has been updated accordingly.

cloudify_cli/tests/commands/test_teardown.py
```diff
@@ -22,6 +24,7 @@ def test_teardown_has_existing_deployments_ignore_deployments(self, mock_teardow
         )
         self.invoke('cfy use 10.0.0.1 -m admin -p admin')
         env.profile = env.get_profile_context(suppress_error=True)
+        sys.modules['cloudify_cli.commands.teardown'].profile = env.profile
```
idanmo: Wouldn't this line effect other tests? perhaps the with patch(...) pattern should be used here?

jcollado: I'm not completely sure. What I can do to make it more consistent with other test cases is modify the import in the teardown module.

jcollado: Fixed in 16690c2

PR #465, opened 2016-10-31
CFY-6009 Add the 'cfy ha start' command for starting a HA cluster

tehasdf:
This commands starts a HA cluster on the current manager, and sets
it as the leader of the cluster.


tehasdf:
depends on #454

tehasdf:
merging back to one PR #454

PR #466, opened 2016-10-31
CFY-6009 Add the 'cfy ha join' command

tehasdf:
Used for adding the current manager to an already existing HA cluster.


tehasdf:
depends on #465

tehasdf:
merging back to one PR #454

PR #485, opened 2016-12-04
CFY-6052 Bootstrap may fail if target for agent key does not exist

yarivgiga:


cloudify_cli/bootstrap/bootstrap.py
```diff
@@ -479,6 +480,10 @@ def _copy_agent_key(agent_local_key_path, agent_remote_key_path,
     if not agent_local_key_path:
         return None
     agent_local_key_path = os.path.expanduser(agent_local_key_path)
+    agent_remote_key_path = os.path.expanduser(agent_remote_key_path)
+    agent_remote_key_dir = '/{0}'.format(
```
dusking: os.path.dirname

cloudify_cli/utils.py
```diff
@@ -91,6 +91,16 @@ def remove_if_exists(path):
             raise  # re-raise exception if a different error occurred
 
 
+def mkdirs(folder_path):
+    try:
+        os.makedirs(folder_path)
```
dusking: - don't try to create dir if exists
- use mkdir command with sudo

PR #489, opened 2016-12-06
CFY-6140 added ldap group dn option

adamlavie:


cloudify_cli/cli/cfy.py
```diff
@@ -594,6 +594,12 @@ def __init__(self):
             required=True,
             help=helptexts.GROUP)
 
+        self.distinguished_name = click.option(
```
dusking: why not ldap_distinguished_name or ldap_group_dn..?

cloudify_cli/commands/user_groups.py
```diff
@@ -50,16 +50,17 @@ def list(sort_by, descending, logger, client):
...
 @cfy.pass_client()
 @cfy.pass_logger
-def create(group_name, logger, client):
+def create(group_name, distinguished_name, logger, client):
     """Create a new group on the manager
 
     `GROUP_NAME` is the name of the new group
     """
-    client.user_groups.create(group_name)
+    client.user_groups.create(group_name, ldap_group_dn=distinguished_name)
```
dusking: do we have default None in case user didn't supply distinguished_name?

PR #490, opened 2016-12-07
CFY-6009 Interface changes in `cfy cluster join/start`

tehasdf:


cloudify_cli/env.py
```diff
@@ -459,20 +460,33 @@ def get_auth_header(username, password):
...
         super(ClusterHTTPClient, self).__init__(*args, **kwargs)
         if not profile.cluster:
             raise ValueError('Cluster client invoked for an empty cluster!')
         self._cluster = list(profile.cluster)
 
     def do_request(self, *args, **kwargs):
-        for node in profile.cluster:
+        copied_data = None
+        if isinstance(kwargs.get('data'), types.GeneratorType):
+            copied_data = itertools.tee(kwargs.pop('data'),
```
tehasdf: this request can be retried for each manager - if the data is a generator, we need to copy it, so we can send it more than once

cloudify_cli/commands/cluster.py
```diff
@@ -131,39 +130,70 @@ def join(client,
...
         'host_ip': cluster_host_ip,
         'node_name': cluster_node_name,
-        'consul_key': cluster_encryption_key,
-        'join_addrs': cluster_join
+        'encryption_key': encryption_key,
+        'join_addrs': join
     })
-    status = _wait_for_cluster_initialized(client, logger)
+    try:
+        status = _wait_for_cluster_initialized(client, logger, timeout=timeout)
```
dusking: wouldn't it be more precise/safe if we'll pass the calculated deadline, instead of the timeout arg?

cloudify_cli/env.py
```diff
@@ -459,20 +460,35 @@ def get_auth_header(username, password):
 
 
 class ClusterHTTPClient(HTTPClient):
+    timeout = 5
```
dusking: is think that default_timeout_sec is more descriptive..

cloudify_cli/env.py
```diff
@@ -459,20 +460,35 @@ def get_auth_header(username, password):
...
-        for node in profile.cluster:
+        # this request can be retried for each manager - if the data is
+        # a generator, we need to copy it, so we can send it more than once
+        copied_data = None
+        if isinstance(kwargs.get('data'), types.GeneratorType):
+            copied_data = itertools.tee(kwargs.pop('data'),
+                                        len(profile.cluster))
+
+        if kwargs.get('timeout') is None:
+            kwargs['timeout'] = self.timeout
```
dusking: instead of last 2 lines, can use: kwargs.setdefault('timeout', self.timeout)
or kwargs.setdefault('timeout', self.default_timeout_sec)

PR #524, opened 2017-02-02
CFY-7103 Pass inputs to start/join to allow configuration

tehasdf:


cloudify_cli/commands/cluster.py
```diff
@@ -92,11 +92,13 @@ def status(client, logger):
 @cfy.pass_client()
 @cfy.pass_logger
 @cfy.options.timeout()
+@cfy.options.inputs
```
tehasdf: change name to `options`

PR #543, opened 2017-03-01
CFY-6502 Add the `cfy cluster set-active` command

tehasdf:


cloudify_cli/commands/cluster.py
```diff
@@ -221,6 +220,31 @@ def update_profile(client,
...
+@cfy.argument('node_name')
+@cfy.pass_client()
+@cfy.pass_logger
+def set_active(client, logger, node_name):
+    nodes = client.cluster.nodes.list()
+    for node in nodes:
+        if node['name'] != node_name:
+            continue
+        if not node['online']:
+            raise CloudifyCliError('{0} is offline!'.format(node_name))
```
dusking: Can't set node {0} as the active, it's offline

cloudify_cli/commands/cluster.py
```diff
@@ -221,6 +220,31 @@ def update_profile(client,
...
+        if node['master']:
+            raise CloudifyCliError('{0} is already the current active node!'
+                                   .format(node_name))
+        break
+    else:
+        raise CloudifyCliError('{0} is not a member of the cluster!'
```
dusking: Can't set node {0} as the active, it's not a member of the cluster

PR #544, opened 2017-03-01
CFY-6503 `cfy teardown`: pass the correct argument to get_rest_client

tehasdf:


cloudify_cli/commands/teardown.py
```diff
@@ -105,7 +105,7 @@ def _update_local_provider_context(ctx, manager_ip, logger):
 
 
 def _get_number_of_deployments(manager_ip):
-    client = env.get_rest_client(manager_ip)
```
tehasdf: just `get_rest_client('blabla')` uses it as the profile object - first argument to `get_rest_client`

cloudify_cli/commands/teardown.py
```diff
@@ -94,7 +94,7 @@ def _update_local_provider_context(ctx, manager_ip, logger):
     try:
         ctx.invoke(
             use,
-            profile_name=manager_ip,
+            manager_ip=manager_ip,
```
tehasdf: otherwise it says `using manager None with port 80`

PR #549, opened 2017-03-05
CFY-6511 Vague error message on cfy ssh

dusking:


cloudify_cli/env.py
```diff
@@ -257,8 +257,8 @@ def get_rest_client(client_profile=None,
 def build_manager_host_string(ssh_user='', ip=''):
     ssh_user = ssh_user or profile.ssh_user
     if not ssh_user:
-        raise CloudifyCliError('Manager User is not set in '
-                               'working directory settings')
+        raise CloudifyCliError('Manager ssh_user is not set '
```
nir0s: would put that in backticks -> `ssh_user` instead of ssh_user

cloudify_cli/env.py
```diff
@@ -257,8 +257,8 @@ def get_rest_client(client_profile=None,
...
+                               'in Cloudify cli settings')
```
nir0s: cli -> CLI

cloudify_cli/ssh.py
```diff
@@ -102,19 +102,19 @@ def execute():
 
 
 def test_profile():
-    msg = 'Manager {0} is not set in working directory settings'
+    msg = 'Manager ssh_{0} is not set in Cloudify cli settings'
```
nir0s: same (backticks)
same (CLI)

PR #556, opened 2017-03-13
Use a timeout by default with a cluster client

tehasdf:
For the failover to work properly, we need to timeout requests
and retry with the other nodes in the cluster


cloudify_cli/env.py
```diff
@@ -505,7 +505,8 @@ def do_request(self, *args, **kwargs):
             copied_data = itertools.tee(kwargs.pop('data'),
                                         len(self._cluster))
 
-        kwargs.setdefault('timeout', self.default_timeout_sec)
```
tehasdf: can't just use `setdefault` - timeout is None usually; but we do need the default timeout to be something short

PR #569, opened 2017-03-26
CFY-6651 prettify secret get

inbala:


inbala:
screen shot 2017-03-26 at 13 06 28

PR #585, opened 2017-05-16
CFY-6870 add certificates replacement after snapshot restore

yarivgiga:


cloudify_cli/cli/helptexts.py
```diff
@@ -217,6 +217,12 @@
                                'automatically. If the tenant name passed ' \
                                'already exists on the manager, an error ' \
                                'will be raised'
+RESTORE_CERTIFICATES = 'Restore the certificates from the snapshot and put ' \
```
dusking: need to get Chana approval..

chanagershom: Restore the certificates from the **snapshot, using them to replace** ~and put them instead of~ the existing Manager certificates. 
If the **certificates'** metadata (i.e. the Manager IP address) from the snapshot does not match the Manager metadata, the certificates will not be restored ~since they would~ **cannot** work on this Manager.

(_Note that all instances of "manager" must be capitalized, when not in the actual code.
Also, I would change ```the existing Manager certificates```, which could be confusing--does it mean the existing one from where you took the snapshot or the existing one on which you are applying the snapshot--to ```the Manager certificates on the newly created Manager.``` or something like that._)

PR #590, opened 2017-05-22
Upgrade & loosen fabric dependency

funkyHat:
currently clashing with cloudify-fabric-plugin

### DIFF
--- a/cloudify_cli/VERSION
+++ b/cloudify_cli/VERSION
@@ -1,4 +1,4 @@
 {
   "edition": "premium",
-  "version": "4.1.0-m2"
+  "version": "4.2.0-.dev1"
 }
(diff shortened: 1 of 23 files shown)


idanmo:
This PR cannot be merged as it breaks the CLI packages build process.
We have previously tried to upgrade and it was too hard getting omnibus to work with a newer version of Fabric.

funkyHat:
Oh dear :( this means the cli is incompatible with the latest `cloudify-fabric-plugin`...

funkyHat:
Could it be `>=1.8.3` instead (assuming there's some way to pin the version in omnibus outside of setup.py)?

idanmo:
The problem was that a newer fabric version requires and newer version of libffi.
After investigating this a little, a newer pip version used when building with omnibus should solve the problem as it bundles all the required dependencies within the final package.
I couldn't find an easy way to specify the latest pip version in omnibus so perhaps you can try yourself..

funkyHat:
I think this should do it

funkyHat:
This was fixed in #631 instead

PR #591, opened 2017-05-24
CFY-6901 When setting the active profile, also set it in-process

tehasdf:
`set_active_profile`, instead of just writing to the active.profile
file, will now set `env.profile = active_profile` as well.

This is done using the global keyword so it is somewhat dirty, but we
are using the profile as a global after all.

This is because bootstrap creates and sets a new profile, so the rest
of bootstrap code should be able to use that profile. Specifically,
without this change, the get_rest_client call will still use the
previous (non-temp, whatever profile was active before) profile,
and if that previous active profile had a cluster, then it'll try
to use the cluster client.

### DIFF
--- a/cloudify_cli/commands/profiles.py
+++ b/cloudify_cli/commands/profiles.py
@@ -262,7 +262,6 @@ def set(profile_name,
                                    .format(profile_name))
         old_name = env.profile.profile_name
         env.profile.profile_name = profile_name
-        env.set_active_profile(profile_name)
     if manager_username:
         logger.info('Setting username to `{0}`'.format(manager_username))
         env.profile.manager_username = manager_username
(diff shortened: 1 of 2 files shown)


tehasdf:
The issue caused by this is that during bootstrap, calls to `get_rest_client` will still use the old profile: https://github.com/cloudify-cosmo/cloudify-cli/blob/472e8811190c8517a1d921b43c1e9af9a135ac8d/cloudify_cli/bootstrap/bootstrap.py#L314
It passes all the inputs explicitly, so they're not taken from the old profile... except for `cluster`, which is :)

Making it actually use the temp profile instead means that we can use `get_rest_client()` without parameters here, they'll be just taken from the profile. And also next time we add a new `get_rest_client` call anywhere in the bootstrap procedure, we won't need to again remember to pass every argument

PR #592, opened 2017-05-24
CFY-6930 Make sure the cluster list is stored in the profile

tehasdf:


cloudify_cli/env.py
```diff
@@ -423,9 +423,11 @@ def profile_name(self, profile_name):
 
     @property
     def cluster(self):
-        # default the .cluster attribute here, so that all callers can use it
-        # as just .cluster, even if it's not present in the source yaml
-        return getattr(self, '_cluster', [])
```
tehasdf: in case there's no `.cluster`, this will just return the empty list and not set it, so if the client code only calls `.append()`, that will be lost!

PR #604, opened 2017-07-02
CFY-7048 Make `profiles use` don't use the old profile for httpscheck

tehasdf:
Pass `profile` explicitly to `get_rest_client`, so that there's no way
this uses the old profile anymore.


tehasdf:
Also fixes CFY-7030

PR #607, opened 2017-07-03
Remove tenant name for snapshot restores

geokala:


cloudify_cli/cli/helptexts.py
```diff
@@ -206,17 +206,6 @@
        'This argument can be used multiple times'
 PERMISSION = 'The permission applicable to a resource [viewer|owner] ' \
              '(default:viewer)'
-RESTORE_SNAPSHOT_TENANT_NAME = 'The name of the tenant into which the ' \
```
idanmo: Perhaps the text should include a description saying that if the snapshot is of a 3.4.X manager the snapshot will be restored to "this" tenant?

geokala: I thought we were removing the entire thing? There is no helptext with it removed.

PR #622, opened 2017-07-18
CFY-7081 When setting verbosity, don't lose the value

tehasdf:
set_cli_except_hook expects the verbosity value, it needs to be
passed down from set_verbosity_level via main.
Otherwise it's just lost.


tehasdf:
Without this, the `verbose` in https://github.com/cloudify-cosmo/cloudify-cli/blob/5119c463679bc29bdb844f2069ecc964e09609a2/cloudify_cli/main.py#L68 is always just `None`

PR #643, opened 2017-08-03
CFY-7128 fixed bug in agents install and added tests

adigrabow:


cloudify_cli/commands/agents.py
```diff
@@ -79,19 +79,29 @@ def install(deployment_id,
     """
     # install agents across all tenants
     if all_tenants:
+        error_flag = True   # True = has errors
```
mcouthon: Maybe find a more descriptive name (like `any_deployments_found`, or `no_deployments_found`

cloudify_cli/commands/agents.py
```diff
@@ -79,19 +79,29 @@ def install(deployment_id,
...
         tenants_list = [tenant.name for tenant in client.tenants.list()]
         for tenant in tenants_list:
```
mcouthon: This is becoming a pretty large function. Maybe it would be better to split the all_tenants case and the single tenant case into 2 separate functions.

cloudify_cli/commands/agents.py
```diff
@@ -109,16 +119,20 @@ def create_deployments_list(client, deployment_id, logger):
     :return A list of the relevant deployments.
             """
     # install agents for a specified deployment
+    raise_error = False
```
mcouthon: This flag is actually not really necessary, as you can decide on whether there was an error or not, just by looking at the error message. If `log_msg` is None, you know there was no error, and if it's anything other than None, you there was. So you can only return 2 values - deps and log_msg

cloudify_cli/tests/commands/test_agents.py
```diff
@@ -232,3 +232,75 @@ def f():
         self.invoke('cfy agents install -t tenant2 dep11',
                     exception=SuppressedCloudifyCliError,
                     err_str_segment='')
+
+    @patch('cloudify_cli.commands.agents.run_worker')
```
mcouthon: Good job on adding tests for every case! 🥇

cloudify_cli/commands/agents.py
funkyHat: If `log_msg` is really `error_msg` could it be named better so that's clear?

cloudify_cli/commands/agents.py
```diff
@@ -79,19 +79,28 @@ def install(deployment_id,
...
+        if no_deployments_found:
+            logger.error(log_msg)
+            raise SuppressedCloudifyCliError()
     else:
         # install agents for all deployments under a specified tenant
         if tenant_name:
             logger.info('Explicitly using tenant `{0}`'.format(tenant_name))
-        deps = create_deployments_list(client, deployment_id, logger)
+        deps, log_msg = create_deployments_list(client, deployment_id, logger)
+        if log_msg:
```
funkyHat: If `log_msg` is really `error_msg` could it be named better?

mcouthon: I agree

PR #661, opened 2017-08-27
CFY-7186: print root cause when YAML parsing error occurs

isaac-s:


cloudify_cli/inputs.py
```diff
@@ -55,8 +55,10 @@ def inputs_to_dict(resources):
                     "- A single quoted wildcard based path "
                     "(e.g. '*-inputs.yaml')\n "
                     "- A string formatted as JSON/YAML\n "
-                    "- A string formatted as key1=value1;key2=value2".format(
-                        resource))
+                    "- A string formatted as key1=value1;key2=value2\n"
+                    "\n"
```
mcouthon: First of all, this can all probably fit in one line.
Second - this seems like something that should be only relevant when running with `-vv` or `-vvv`, not always. I would log it in debug, instead of adding it to the error message.

isaac-s: @mcouthon I added the `\n` on a separate line for clarity, but I can revert that.
Regarding `-vv`, that would be IMO missing the point. 

1. 100% of our customers/users, especially beginners, run `cfy` commands without `-vv`. 
2. If we insist to have the root reason only printed with `-vv`, then the "ordinary" message (without `-vv`) should at least give the user some hint as to what's wrong, and tell them to use `-vv`.

I strongly prefer option 1 because in this case it's much more intuitive.

mcouthon: Can you give me an example of what such output will look like? I'm not sure I understand what's the added value here.

isaac-s: @mcouthon example: `-i` points to a misformatted YAML file.

Currently, the error you'll get:

```
Invalid input: me.yaml. It must represent a dictionary. Valid values can be one of:
 - A path to a YAML file
 - A path to a directory containing YAML files
 - A single quoted wildcard based path (e.g. '*-inputs.yaml')
 - A string formatted as JSON/YAML
 - A string formatted as key1=value1;key2=value2
```

With my fix, you'll get:

```
Invalid input: me.yaml. It must represent a dictionary. Valid values can be one of:
 - A path to a YAML file
 - A path to a directory containing YAML files
 - A single quoted wildcard based path (e.g. '*-inputs.yaml')
 - A string formatted as JSON/YAML
 - A string formatted as key1=value1;key2=value2

Root cause: <some text that shows you precisely what's wrong with your YAML
```

Looks like huge added value to me.

mcouthon: Ok, this looks benign enough. Green

mcouthon: One thing though - please add a unittest, to make sure you're getting correct error messages. Just a couple of malformed inputs.

isaac-s:
@mcouthon added tests and merged.

PR #662, opened 2017-08-27
CFY-7184 Add all_tenants option to executions list

inbala:


cloudify_cli/commands/executions.py
```diff
@@ -88,6 +88,7 @@ def manager_get(execution_id, logger, client, tenant_name):
 @cfy.options.sort_by()
 @cfy.options.descending
 @cfy.options.tenant_name(required=False, resource_name_for_help='executions')
```
yarivgiga: You have to change it to "tenant_name_for_list" (which is mutually exclusive with all_tenants).
Check out blueprints/deployments list and do the same.

PR #673, opened 2017-09-18
CFY-7253: remove ssl redirect

isaac-s:
* Added `--secured` switch to `cfy profiles use`
* Removed dummy `status` request for the purpose of determining protocol & port
* Protocol is `https` if `--secured` is specified, `http` otherwise
* If `--rest-port` is provided, it is used. Otherwise, the port is `443` if the protocol is `https`, or `80` if the protocol is `http`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,4 +65,5 @@ docs/_build/
 .DS_Store
 
 .idea/
+.cloudify/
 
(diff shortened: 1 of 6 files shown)


isaac-s:
What I did here:

* Removed `--rest-port` from `cfy profiles use` and replaced it with `--secured` (defaults to `false`). More intuitive.
* Removed the dummy `status` request that will *always* fail and was *only* needed for the purpose of the redirect
* Removed the fallback to SSL

Long story short: user must know in advance whether they're connecting to a secured manager or a non-secure one. No guesswork, less bugs.

PR #674, opened 2017-09-22
CFY-6946: omit -e for 'cfy events list', as execution-id is mandatory

isaac-s:


isaac-s:
@tehasdf what about now?

PR #676, opened 2017-09-24
CFY-6850: simplified error reporting when missing / unknown inputs are provided

isaac-s:
In case of missing input:

```
Unable to create deployment. Not all required inputs have been specified. Missing input(s): <comma-separated-list-of-inputs>
```

In case of unknown input:

```
Unable to create deployment, unknown input(s) were specified: <comma-separated-list-of-inputs>
```

The CLI exception itself is suppressed from the console log, but _is_ printed to the CLI log file.

### DIFF
--- a/cloudify_cli/commands/deployments.py
+++ b/cloudify_cli/commands/deployments.py
@@ -46,13 +46,17 @@ def deployments():
     pass
 
 
+def _get_deployment_inputs(client, blueprint_id):
+    blueprint = client.blueprints.get(blueprint_id)
+    return blueprint.plan['inputs']
+
+
 @cfy.pass_logger
 def _print_deployment_inputs(client, blueprint_id, logger):
-    blueprint = client.blueprints.get(blueprint_id)
-
     logger.info('Deployment inputs:')
     inputs_output = StringIO()
-    for input_name, input_def in blueprint.plan['inputs'].iteritems():
+    for input_name, input_def in _get_deployment_inputs(
+            client, blueprint_id).iteritems():
         inputs_output.write('\t{0}:{1}'.format(input_name, os.linesep))
         for k, v in input_def.iteritems():
             inputs_output.write('\t\t{0}: {1}{2}'.format(k, v, os.linesep))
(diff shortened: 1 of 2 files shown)


cloudify_cli/commands/deployments.py
```diff
@@ -227,10 +231,16 @@ def manager_create(blueprint_id,
             skip_plugins_validation=skip_plugins_validation
         )
     except MissingRequiredDeploymentInputError as e:
-        logger.info('Unable to create deployment. Not all '
-                    'required inputs have been specified...')
-        _print_deployment_inputs(client, blueprint_id)
-        raise CloudifyCliError(str(e))
+        expected_inputs_def = _get_deployment_inputs(client, blueprint_id)
```
mcouthon: TBH, I don't much like that there are calculations inside an `except` clause. If something isn't right there, the error won't be caught, and will be even less helpful.

cloudify_cli/commands/deployments.py
```diff
@@ -227,10 +231,16 @@ def manager_create(blueprint_id,
...
+        available_input_names = set(inputs.keys())
```
mcouthon: And in any case, this logic should be done in the DSL parser, or on the server side, anyway. We're trying to move the CLI towards being a rather stupid conduit for REST commands.

cloudify_cli/tests/commands/test_deployments.py
```diff
@@ -360,21 +367,26 @@ def raise_error(*args, **kwargs):
...
         for output in expected_outputs:
-            self.assertIn(output, outcome)
-
-        self.assertIn(error_msg, outcome)
+            found = False
+            for outcome_line in outcome:
+                if output in outcome_line:
+                    found = True
+                    break
+            self.assertTrue(found, 'String ''{0}'' not found in outcome {1}'
```
mcouthon: Are those double single quotes?

cloudify_cli/commands/deployments.py
geokala: Surely we still need to handle missing input stuff on the client side anyway, for cfy local cases if not to speed up returning of useful errors to a user rather than requiring a roundtrip to the manager?

cloudify_cli/commands/deployments.py
isaac-s: @mcouthon I don't like it either, but this is the existing situation too. As of 4.1.1, we make a call to the manager to retrieve the blueprint in order to print the deployment inputs.

cloudify_cli/tests/commands/test_deployments.py
isaac-s: Yes

cloudify_cli/commands/deployments.py
isaac-s: @mcouthon @geokala all the info required to get these details, is already available on the manager side.
At the time of throwing the exception on the manager side, we have all the info. Unfortunately, our Exception classes are "flat" and don't contain fine-grained fields, so we can't pass this info back from the manager to the CLI in an easy way (all we have at the CLI side is just a flat string explaining what's wrong).

Alternatively what we could do is simply print out the error message received from the REST service and remove `_print_deployment_inputs` altogether.

isaac-s:
@mcouthon @geokala See the recent commit. Much simpler fix, which also removes the call to the manager which currently exists in 4.1.1.

PR #696, opened 2017-10-24
CFY-7199 Work with 4.1.1 (and earlier) profiles

geokala:


geokala:
4.1.x -> 4.2+ is already broken, closing.

PR #697, opened 2017-10-25
CFY-7256 verify role is given when adding user to tenant

dusking:


cloudify_cli/commands/tenants.py
```diff
@@ -71,7 +71,7 @@ def create(tenant_name, logger, client):
 @tenants.command(name='add-user',
                  short_help='Add a user to a tenant [manager only]')
 @cfy.argument('username', callback=cfy.validate_name)
-@cfy.options.user_tenant_role()
+@cfy.options.user_tenant_role(required=True)
```
jcollado: Since `required` will always be set to `True`. I suggest to update `cfy.options.user_tenant_role` to remove the `required` parameter and hardcode its value to `True` there. Also, since no other parameters are there, convert it from a `staticmethod` to a `property`.

PR #699, opened 2017-10-26
CFY-6680: upgrade to click==4.1

isaac-s:
The latest is 6.x, but I chose 4.1 (the latest in the 4.x family) in order to reduce the risk of backward compatibility issues.


tehasdf:
done in #784

PR #700, opened 2017-10-30
CFY-7442: Add skip plugins validation

isaac-s:


cloudify_cli/commands/plugins.py
```diff
@@ -100,7 +102,8 @@ def upload(ctx, plugin_path, private_resource, logger, client, tenant_name):
     if tenant_name:
         logger.info('Explicitly using tenant `{0}`'.format(tenant_name))
     parsed_url = urlparse(plugin_path)
-    if not parsed_url.scheme or not parsed_url.netloc:
+    if (not parsed_url.scheme or not parsed_url.netloc) and not \
```
tehasdf: how about factoring this condition out so that it's easier to understand/self-documenting - something like `is_local = not parsed_url.scheme or not parsed_url.netloc`?

cloudify_cli/cli/cfy.py
```diff
@@ -803,6 +803,13 @@ def __init__(self):
             help=helptexts.SKIP_PLUGINS_VALIDATION
         )
 
+        self.skip_plugins_upload_validation = click.option(
+            '--skip-validation',
```
tehasdf: this only affects local plugins - maybe it should be reflected in the flag name?

isaac-s: `--skip-local-plugins-validation`? would be too long.
I chose a flag name that makes sense to the user.

isaac-s: However, the help text should probably be adjusted. Will fix.

PR #701, opened 2017-11-01
CFY-7439 cfy profiles use: only switch, don't update existing profiles

tehasdf:
For already existing profiles, only switch to them, don't actually try
to get provider context, or update the profile.

currently, `cfy profiles use` has 3 modes of working:
  1. create a new profile
  2. switch profiles
  3. switch profiles AND update the existing profile, 
      eg `cfy profiles use existing.profile.ip --password updated_password`

This change gets rid of 3, so it is an error to pass any flags (like `--password`)
when using `cfy profiles use` with an already existing profile.


tehasdf:
In the future, we should probably split creating and switching profiles to separate subcommands.

PR #703, opened 2017-11-02
CFY-7439 Don't update provider context in teardown

tehasdf:
This was related to the recovery process, which isn't supported
anymore.


tehasdf:
todo test with a manager that didn't finish bootstrapping correctly (--keep-up-on-failure)
cf https://cloudifysource.atlassian.net/browse/CFY-1480

PR #705, opened 2017-11-06
CFY-7332 Add a command to `profiles set` for a specific cluster node

tehasdf:
Similar to how `cfy profiles set` allows setting things like the
REST cert and ssh params for a profile, `cfy profiles set-cluster`
allows setting those for a specific node in a cluster.

This means we can use the cli to set a different cert for every
cluster node.

### DIFF
--- a/cloudify_cli/commands/profiles.py
+++ b/cloudify_cli/commands/profiles.py
@@ -387,7 +387,7 @@ def set_cmd(profile_name,
 @profiles.command(
     name='set-cluster',
     short_help='Set connection options for a cluster node')
-@cfy.options.cluster_node_name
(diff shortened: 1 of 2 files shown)


cloudify_cli/commands/profiles.py
```diff
@@ -385,6 +385,52 @@ def set_cmd(profile_name,
 
 
 @profiles.command(
+    name='set-cluster',
+    short_help='Set connection options for a cluster node')
+@cfy.options.cluster_node_name
```
mcouthon: Why is this an option and not an argument?

tehasdf: good point, thanks

PR #708, opened 2017-11-21
CFY-7554 Add commands for setting the availability of a resource

inbala:


cloudify_cli/cli/helptexts.py
```diff
@@ -244,3 +242,11 @@
 SECRET_FILE = "The secret's file to use its content as value to be set"
 CLUSTER_NODE_OPTIONS = 'Additional options for the cluster node '\
                        'configuration {0}'.format(INPUTS_PARAMS_USAGE)
+TENANT_AVAILABILITY = "Sets the resource's availability so the resource " \
```
dusking: TENANT_AVAILABILITY = "Makes the resource available to all users that " \
                      "have access to the same tenant. You cannot use " \
                      "this argument with arguments: [global_availability] "
GLOBAL_AVAILABILITY = "Makes the resource available to all users that " \
                      "have access to at least one of the manager's " \
                      "tenants. You cannot use this argument with " \
                      "arguments: [tennant_availability] "

PR #710, opened 2017-11-23
CFY-7562. Secret create update_if_exists

jcollado:
In this PR, the create secret command is updated to handle the upsert flag that is now supported by the REST API.

Requires: cloudify-cosmo/cloudify-rest-client#206

### DIFF
--- a/cloudify_cli/cli/helptexts.py
+++ b/cloudify_cli/cli/helptexts.py
@@ -236,6 +236,7 @@ PROFILE_NAME = 'Name of the profile to use'
 SECRET_VALUE = "The secret's value to be set"
 SECRET_STRING = "The string to use as the secret's value"
 SECRET_FILE = "The secret's file to use its content as value to be set"
+SECRET_UPSERT = 'Update secret with new value if it already exists'
 CLUSTER_NODE_OPTIONS = 'Additional options for the cluster node '\
                        'configuration {0}'.format(INPUTS_PARAMS_USAGE)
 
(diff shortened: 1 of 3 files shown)


funkyHat:
Why are we calling this option `--upsert` and not something more obvious like `--update`? (UPSERT is only likely to be familiar if you've encountered it in the context of RDBMSs... I hadn't heard of it before today despite having used databases for ~ 15 years...)

cloudify_cli/cli/helptexts.py
```diff
@@ -242,5 +242,6 @@
 SECRET_VALUE = "The secret's value to be set"
 SECRET_STRING = "The string to use as the secret's value"
 SECRET_FILE = "The secret's file to use its content as value to be set"
+SECRET_UPSERT = 'Update secret with new value if it already exists'
```
geokala: Do we not mean: "Update secret with new value if it exists, otherwise create it."?

geokala:
@funkyHat I think 'upsert' is reasonable as an arg name here, as long as our helptext is accurate. Our current helptext suggests that it really is just update, not upsert.

cloudify_cli/cli/helptexts.py
geokala: (I realise this is on a create command so create is somewhat implicit, but this feels like it might be better explicit)

jcollado:
@funkyHat @geokala Thanks for your review. `--upsert` has been renamed to `--update-if-exists` which indeed should be clearer.

PR #717, opened 2017-12-12
CFY-7662: if --rest-certificate is provided, automatically assume SSL

isaac-s:


cloudify_cli/commands/profiles.py
```diff
@@ -182,6 +182,8 @@ def _create_profile(
         rest_certificate,
         skip_credentials_validation,
         logger):
+    ssl = ssl or rest_certificate
```
tehasdf: this of course works, but i think i'd prefer to keep it strictly bools, eg.
```python
if rest_certificate:
    ssl = True
```
because i think that makes the intent more obvious, and i'm worried that in the future some unrelated part of code might eg. start storing `ssl` in the context

PR #725, opened 2017-12-26
CFY-7689 Not all credentials passed to `cfy profiles set` are validated

yarivgiga:


cloudify_cli/tests/commands/test_profiles.py
```diff
@@ -292,7 +292,7 @@ def test_set_works_without_skip(self, validate_credentials_mock):
         self.use_manager()
         self.invoke('profiles set -u 0 -p 0 -t 0')
 
-        validate_credentials_mock.assert_called_once_with('0', '0', '0')
+        validate_credentials_mock.assert_called_once_with('0', '0', '0', None)
```
mcouthon: Maybe instead of passing None, let's also check passing `-c`? :)

cloudify_cli/commands/profiles.py
```diff
@@ -478,7 +478,7 @@ def unset(manager_username,
         tenant = env.profile.manager_tenant
 
     if not skip_credentials_validation:
-        _validate_credentials(username, password, tenant)
+        _validate_credentials(username, password, tenant, None)
```
mcouthon: Let's add a small comment here, to explain why we're passing `None`

PR #732, opened 2018-01-17
CFY-7800 Support batch plugin installation

mxmrlv:


cloudify_cli/commands/plugins.py
```diff
@@ -127,6 +163,32 @@ def upload(ctx,
...
+            open(utils.get_local_path(plugin_mappings)))
+    except CloudifyCliError:
+        plugin_mappings = yaml.load(plugin_mappings)
+
+    cvn_path = _create_caravan(plugin_mappings, destination, name)
+    logger.info('Caravan created at {0}'.format(cvn_path))
+    return cvn_path
+
+
+@plugins.command(name='create-caravan', short_help='Create a caravan')
```
idob-cloudify: need to be 'upload-caravan'

PR #734, opened 2018-01-18
CFY-7801 Error message for using `plugin:` with `blueprint validate`

tehasdf:
This blueprint can't be validated locally, instead it must be
uploaded to a manager.

Example:
```
Validating blueprint: bp.yaml
Error fetching plugin yaml: 'plugin:cloudify-openstack-plugin'
Blueprints using plugin repository imports can not be validated locally.
Possible solutions:
  - Upload the blueprint to a Cloudify Manager
  - Use an explicit URL to the plugin YAML file instead of a plugin repository `plugin:` import.

```

### DIFF
--- a/cloudify_cli/config/config.py
+++ b/cloudify_cli/config/config.py
@@ -20,8 +20,11 @@ import yaml
 
 from dsl_parser import utils as dsl_parser_utils
 from dsl_parser.constants import IMPORT_RESOLVER_KEY
+from dsl_parser.import_resolver.default_import_resolver import (
+    DefaultImportResolver
+)
 
-from .. import env
+from .. import env, exceptions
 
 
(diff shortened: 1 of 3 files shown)


cloudify_cli/config/config.py
```diff
@@ -102,3 +107,25 @@ def is_validate_definitions_version():
...
+        if self._is_plugin_url(import_url):
+            e = exceptions.CloudifyCliError(
+                'Error fetching plugin yaml: {0!r}\nBlueprints using plugin '
+                'repository imports can not be validated locally.'
+                .format(import_url))
+            e.possible_solutions = [
+                'You can use an explicit URL to the plugin YAML to perform '
+                'local validation.',
+                'Uploading the blueprint to the manager will also run the '
+                'validation process.'
```
dusking: ..will run the validation process `, and supports also the "plugin:" notation.`

PR #735, opened 2018-01-18
CFY-6732 add ldap get command

yarivgiga:


cloudify_cli/commands/ldap.py
```diff
@@ -51,3 +51,11 @@ def set(ldap_server,
                     ldap_domain=ldap_domain,
                     ldap_dn_extra=ldap_dn_extra)
     logger.info('LDAP authentication set successfully')
+
+
+@ldap.command(name='get',
```
dusking: why not status?

PR #747, opened 2018-01-23
work around pip extras bug

funkyHat:
https://github.com/pypa/pip/issues/4957


funkyHat:
This was worked around a different way in #749

PR #755, opened 2018-01-28
CFY-7835 fix local plugin upload

dusking:


cloudify_cli/utils.py
```diff
@@ -325,8 +325,8 @@ def get_local_path(source, destination=None, create_temp=False):
         return downloaded_file
     elif os.path.isfile(source):
         if not destination and create_temp:
-            fd, destination = tempfile.mkstemp()
-            os.close(fd)
+            source_name = os.path.basename(source)
```
mxmrlv: why not just remove the create_temp from the function and pass a temporary dir to destination?

dusking: yeah, it's the same.. just because I'm using it multiple times..

PR #774, opened 2018-02-23
build script: update nss

tehasdf:
On centos, nss (the ssl/tls lib used by curl & git) must be a recent
enough version to support tlsv1.2, so that we can connect to github.
As of 23.02.2018, github has dropped support for old tls versions
( https://github.com/blog/2507-weak-cryptographic-standards-removed )


packaging/linux/provision.sh
```diff
@@ -51,17 +51,18 @@ function prepare_linux () {
...
         sudo yum install -y http://opensource.wandisco.com/centos/6/git/x86_64/wandisco-git-release-6-1.noarch.rpm
         sudo yum install -y git fakeroot python-devel rpm-build
+        sudo yum update -y nss
         gpg=gpg2
     else
         sudo apt-get install -y git curl fakeroot python-dev
         gpg=gpg
     fi
-    
+
```
geokala: Well, that's my hidden whitespace program broken!

PR #821, opened 2018-05-27
CY-372-cli-download-cluster-logs

idob-cloudify:


cloudify_cli/commands/logs.py
```diff
@@ -62,12 +67,33 @@ def _archive_logs(logger):
 def download(output_path, logger):
     """Download an archive containing all of the manager's service logs
     """
-    archive_path_on_manager = _archive_logs(logger)
-    logger.info('Downloading archive to: {0}'.format(output_path))
-    ssh.get_file_from_manager(archive_path_on_manager, output_path)
-    logger.info('Removing archive from manager...')
-    ssh.run_command_on_manager(
-        'rm {0}'.format(archive_path_on_manager), use_sudo=True)
+    if len(env.profile.cluster) > 0:
```
tehasdf: better just `if env.profile.cluster` in case it gets set to none

PR #831, opened 2018-06-09
CY-383 : Adding the option to create a user and associating with a tenant and user role

tyacbovi:
Also fixed the help msg of security role flag.

### DIFF
--- a/cloudify_cli/cli/cfy.py
+++ b/cloudify_cli/cli/cfy.py
@@ -1141,18 +1141,20 @@ class Options(object):
             help=helptexts.BLUEPRINT_PATH)
 
     @staticmethod
-    def tenant_role(help_text, required=False):
-        return click.option(
-            '-r',
-            '--role',
-            required=required,
-            help=help_text
-        )
+    def tenant_role(help_text, required, options_flags=None):
+        args = options_flags or ['-r', '--role']
+
+        kwargs = {
+            'required': required,
+            'help': help_text
+        }
+        return click.option(*args, **kwargs)
 
     @staticmethod
-    def user_tenant_role():
+    def user_tenant_role(required=True, options_flags=None):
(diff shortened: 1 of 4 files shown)


cloudify_cli/cli/cfy.py
```diff
@@ -1141,18 +1141,22 @@ def blueprint_path(required=False):
...
-        return click.option(
-            '-r',
-            '--role',
-            required=required,
-            help=help_text
-        )
+    def tenant_role(help_text, required, options_flags=None):
+        args = ['-r', '--role']
+        if options_flags:
+            args = options_flags
```
yarivgiga: could be done in one line:
args = option_flags or ['-r', '--role']

cloudify_cli/commands/users.py
```diff
@@ -113,17 +113,27 @@ def list(sort_by,
...
 @cfy.options.password
+@cfy.options.tenant_name(required=False)
+@cfy.options.user_tenant_role(required=False,
+                              options_flags=['-l', '--user-tenant-role'])
 @cfy.assert_manager_active()
 @cfy.pass_client()
 @cfy.pass_logger
-def create(username, security_role, password, logger, client):
+def create(username, security_role, password, tenant_name, user_tenant_role,
+           logger, client):
```
yarivgiga: if func definition is too long for 1 line, better look like this:
func(arg1,
        arg2,
        arg3):

cloudify_cli/commands/users.py
```diff
@@ -113,17 +113,27 @@ def list(sort_by,
...
     """Create a new user on the manager
 
-    `USERNAME` is the username of the user
+     `USERNAME` is the username of the user
```
yarivgiga: why?

PR #834, opened 2018-06-12
CY-384 Add a flag to kill-cancel executions

tehasdf:


cloudify_cli/cli/helptexts.py
```diff
@@ -108,6 +108,9 @@
     "Terminate the execution abruptly, rather than request an orderly "
     "termination"
 )
+KILL_EXECUTION = (
+    "Terminate the execution abruptly, and also stop currently running tasks"
```
mcouthon: Maybe add a word about the fact that we're actually killing user processes here, and not just Cloudify's internal executions?

cloudify_cli/commands/executions.py
```diff
@@ -281,20 +281,26 @@ def manager_start(workflow_id,
...
-        'Force-' if force else '', execution_id))
-    client.executions.cancel(execution_id, force)
+    if kill:
+        message = 'Killing'
+    elif force:
+        message = 'Force-cancelling'
+    else:
+        message = 'Cancelling'
+    logger.info('{0} execution {1}'.format(message, execution_id))
+    client.executions.cancel(execution_id, force=force, kill=kill)
```
mcouthon: This means that there's a change in the rest client, but you didn't send me a PR about it (and I don't see a branch on the topic).

tehasdf: rest-client lives in common now! :P
https://github.com/cloudify-cosmo/cloudify-common/pull/44/files#diff-c1239f5a046a14346e86ca0029acfd70R217

mcouthon: That's... embarrassing :)

PR #837, opened 2018-06-13
CY-408 cluster update-profile: use .public_ip if exists, then .host_ip

tehasdf:


dev-requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/cloudify-cosmo/cloudify-common@master#egg=cloudify-common==4.4.dev1
+git+https://github.com/cloudify-cosmo/cloudify-common@cy-408-public-ip#egg=cloudify-common==4.4.dev1
```
tehasdf: remember to remove

PR #838, opened 2018-06-17
CY-414 : Adding to log message the chosen tenant rule

tyacbovi:
Also adding it to 'add user to tenant' command

### DIFF
--- a/cloudify_cli/commands/tenants.py
+++ b/cloudify_cli/commands/tenants.py
@@ -155,8 +155,8 @@ def add_user(username, tenant_name, role, logger, client):
     with handle_client_error(409, graceful_msg, logger):
         client.tenants.add_user(username, tenant_name, role)
         logger.info(
-            'User `{0}` added successfully to tenant `{1}`'
-            .format(username, tenant_name)
+            'User `{0}` added successfully to tenant `{1}` with `{2}` role'
+            .format(username, tenant_name, role)
         )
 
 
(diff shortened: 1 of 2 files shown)


cloudify_cli/commands/tenants.py
```diff
@@ -155,8 +155,8 @@ def add_user(username, tenant_name, role, logger, client):
     with handle_client_error(409, graceful_msg, logger):
         client.tenants.add_user(username, tenant_name, role)
         logger.info(
-            'User `{0}` added successfully to tenant `{1}`'
-            .format(username, tenant_name)
+            'User `{0}` added successfully to tenant `{1}` in `{2}` rule'
```
yarivgiga: with the role*

cloudify_cli/commands/users.py
```diff
@@ -131,13 +131,14 @@ def create(username,
...
-    logger.info('User `{0}` created'.format(username))
+    logger.info('User `{0}` created with `{1}` security rule'.format(
+        username, security_role))
 
     if tenant_name and user_tenant_role:
         client.tenants.add_user(username, tenant_name, user_tenant_role)
         logger.info(
-            'User `{0}` added successfully to tenant `{1}`'
-            .format(username, tenant_name))
+            'User `{0}` added successfully to tenant `{1}` in `{2}` rule'
```
yarivgiga: with the role*

PR #850, opened 2018-07-16
Added --ignore-plugin-installation-failure flag to 'cfy snapshot rest…

barucoh:
…ore' command, will ignore any plugins failing to install during a snapshot restore operation.


cloudify_cli/cli/helptexts.py
```diff
@@ -282,3 +282,7 @@
                  '(that is connected to the old Cloudify Manager) will be ' \
                  'stopped. *IMPORTANT* if the deployment has monitoring ' \
                  'with auto-healing configured, you need to disable it first'
+IGNORE_PLUGIN_INSTALLATION_FAILURE = 'If set, any plugin errors arising ' \
```
adigrabow: Would change to: if set, plugin installation errors during `snapshot restore` will only be logged as warnings, and will not fail the `snapshot restore` workflow.

barucoh: Fixed

PR #862, opened 2018-08-15
Cy 538 queued executions

adigrabow:


cloudify_cli/commands/executions.py
```diff
@@ -189,11 +192,8 @@ def manager_start(workflow_id,
     events_message = "* Run 'cfy events list -e {0}' to retrieve the " \
                      "execution's events/logs"
     original_timeout = timeout
-    logger.info('Executing workflow {0} on deployment {1} '
-                '[timeout={2} seconds]'.format(
-                    workflow_id,
-                    deployment_id,
-                    timeout))
+    log_msg = 'Executing workflow `{0}` on deployment `{1}` [timeout={2}' \
+              ' seconds]'.format(workflow_id, deployment_id, timeout)
```
tehasdf: i think let's still log BEFORE actually running the execution... and then another log if queued?

cloudify_cli/commands/executions.py
```diff
@@ -232,8 +233,16 @@ def manager_start(workflow_id,
...
+                force=force,
+                queue=queue)
 
+        if execution.status == 'queued':  # We don't need to wait for execution
+            log_msg = 'Execution workflow `{0}` on deployment `{1}` is' \
+                      ' queued. It will start when possible.' \
+                      ''.format(workflow_id, deployment_id)
+            logger.info(log_msg)
+            return
+        logger.info(log_msg)
```
idob-cloudify: why do you need log_msg defined up there is used only here?

cloudify_cli/commands/snapshots.py
```diff
@@ -112,9 +114,14 @@ def create(snapshot_id,
                                         include_metrics,
                                         not exclude_credentials,
                                         not exclude_logs,
-                                        not exclude_events)
-    logger.info("Started workflow execution. The execution's id is {0}".format(
-        execution.id))
+                                        not exclude_events,
+                                        queue)
+    log_msg = "Started workflow execution. The execution's id is {0}".format(
```
idob-cloudify: why not use "if queue.... else: log_msg=..."

cloudify_cli/commands/executions.py
adigrabow: Yes, makes sense

PR #868, opened 2018-09-04
CY-595 factor get_kerberos_indication func into cloudify-common

yarivgiga:


dev-requirements.txt
```diff
@@ -1 +1 @@
-git+https://github.com/cloudify-cosmo/cloudify-common@master#egg=cloudify-common==4.5.dev1
+git+https://github.com/cloudify-cosmo/cloudify-common@CY-595-Enable-setting-kerberos-env-in-mgmtworker#egg=cloudify-common==4.5.dev1
```
yarivgiga: revert before merge

PR #869, opened 2018-09-20
CYBL-582: print operations regardless of verbosity

isaac-s:


tehasdf:
and here: https://github.com/cloudify-cosmo/cloudify-cli/blob/a0d543091eb6a89df6527caddd83ae9aab724985/cloudify_cli/commands/nodes.py#L81

isaac-s:
@tehasdf I updated the JSON mode code as well

PR #875, opened 2018-10-08
CY-682 allow setting ldap without supplying admin credentials

yarivgiga:


cloudify_cli/commands/ldap.py
```diff
@@ -43,6 +45,10 @@ def set(ldap_server,
         ldap_dn_extra,
         client,
         logger):
+    if (ldap_username and not ldap_password) \
```
geokala: None check here would be good too I think (see related cloudify-manager PR for rationale).

PR #878, opened 2018-10-21
CY-693: prevention of imported catalog blueprints

tyacbovi:


dev-requirements.txt
```diff
@@ -1 +1 @@
-git+https://github.com/cloudify-cosmo/cloudify-common@master#egg=cloudify-common==4.5.5.dev1
+git+https://github.com/cloudify-cosmo/cloudify-common@CY-693#egg=cloudify-common==4.5.5.dev1
```
tehasdf: remember

cloudify_cli/tests/commands/test_blueprints.py
```diff
@@ -64,6 +80,14 @@ def test_blueprints_delete(self):
         self.client.blueprints.delete = MagicMock()
         self.invoke('blueprints delete a-blueprint-id')
 
+    def test_blueprints_delete_force(self):
+        for flag in ['--force', '-f']:
+            mock = MagicMock()
```
tehasdf: i guess could have been just a Mock, not magic

cloudify_cli/tests/commands/test_blueprints.py
```diff
@@ -64,6 +80,14 @@ def test_blueprints_delete(self):
...
+            self.client.blueprints.delete = mock
+            self.invoke('cfy blueprints delete a-blueprint-id {0}'.format(
```
tehasdf: this is all over the tests and at some point we should decide one way or another
but currently the `cfy` is not needed here
see https://github.com/cloudify-cosmo/cloudify-cli/blob/master/cloudify_cli/tests/cfy.py#L68

PR #881, opened 2018-10-29
Adjusted response to indicate database location instead of FAIL in ca…

barucoh:
…se of remote database

### DIFF
--- a/cloudify_cli/commands/cluster.py
+++ b/cloudify_cli/commands/cluster.py
@@ -335,6 +335,9 @@ def _prepare_node(node):
     checks = node.pop('checks', {})
     checks = {check: 'OK' if passing else 'FAIL'
               for check, passing in checks.items()}
+    remote_database = node.pop('remote_database', False)
+    if remote_database:
+        checks['database'] = 'REMOTE'
     node.update(checks)
     online = node.pop('online', False)
     master = node.pop('master', False)


cloudify_cli/commands/cluster.py
```diff
@@ -335,6 +335,9 @@ def _prepare_node(node):
     checks = node.pop('checks', {})
     checks = {check: 'OK' if passing else 'FAIL'
               for check, passing in checks.items()}
+    remote_database = node.pop('remote_database', False)
+    if remote_database:
+        checks['database'] = 'REMOTE'
```
mcouthon: I think there are checks somewhere that fail if the statuses are not `OK`. Did you catch those?

barucoh: Clustering with a remote database is a new feature (this one :)), so I don't think there are any checks for that, the current behavior will stay the same in an AIO architecture.
Nonetheless, I checked if this is addressed somewhere and will run it against the current tests to verify.

barucoh: I couldn't find any checks for that, will run our current tests against it to verify.

tehasdf: the value returned from the restservice is just a boolean, we do check for that in system-tests https://github.com/cloudify-cosmo/cloudify-system-tests/blob/master/cosmo_tester/test_suites/ha/ha_helper.py#L41
the string displayed by the cli (OK/FAIL/REMOTE), i dont think we check that

PR #886, opened 2018-12-04
CYBL-626 Add summaries to CLI

geokala:


cloudify_cli/commands/summary.py
```diff
@@ -0,0 +1,104 @@
...
+@cfy.assert_manager_active()
+def summary():
+    """Get summary information
+    """
+    pass
+
+
+@summary.command(name='nodes',
+                 short_help='Retrieve summary of node details [manager only]')
+@cfy.argument('target_field')
```
tehasdf: how about `@argument(..., type=click.choice(['deployment_id']))`

geokala: What does that gain us? I'm guessing it doesn't actually look up deployment IDs? (also, for nodes this might work currently but for node instances it can be deployment_id or node_Id, and we could end up adding other summaries to nodes and deployments later)

tehasdf: validation that only the string "deployment_id" can be passed there and not other things. I mean yes, for node instances the Choice should contain more choices :P

geokala: Ahh, OK- I'll update that tomorrow morning

PR #887, opened 2018-12-09
CY-873 cfy profiles show: show cluster nodes

tehasdf:
Instead of showing just the profile data which contains a single
manager_ip, if the current profile is a cluster profile, show
all stored nodes

### DIFF
--- a/cloudify_cli/commands/profiles.py
+++ b/cloudify_cli/commands/profiles.py
@@ -71,7 +71,24 @@ def show(logger):
         return
 
     active_profile = _get_profile(env.get_active_profile())
-    print_single(PROFILE_COLUMNS, active_profile, 'Active profile:')
+    if active_profile.get('cluster'):
+        common_attributes = {k: active_profile.get(k) for k in PROFILE_COLUMNS}
+        nodes = []
+        for node in active_profile['cluster']:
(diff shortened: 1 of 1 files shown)


tehasdf:
non-cluster output is unchanged, of course

cloudify_cli/commands/profiles.py
```diff
@@ -71,7 +71,24 @@ def show(logger):
...
+        nodes = []
+        for node in active_profile['cluster']:
+            # merge the common attrs with node data, but rename node's name
+            # attribute to cluster_node, because the attribute 'name' is
+            # reserved for the profile name
+            node_data = dict(node)
+            node_data['cluster_node_name'] = node_data.pop('name')
+            nodes.append(dict(common_attributes, **node_data))
+
+        columns = PROFILE_COLUMNS[:1] + ['cluster_node_name'] \
```
tyacbovi: I think PROFILE_COLUMNS[:1] is better as PROFILE_COLUMNS[0], and maybe export into a function for readability

tehasdf: has to be a list, so it'd have to be `[PROFILE_COLUMNS[0]]`, so i thought `[:1]` is better because of the symmetry with `[1:]`. I guess might as well factor out to a function

PR #897, opened 2018-12-26
CY-901 cfy local: add `executions list` and `get`

tehasdf:


tehasdf:
requires #896, will change the base branch to master after that is merged

PR #900, opened 2019-01-01
Validate scheduled date, and convert to UTC in CLI

adigrabow:


cloudify_cli/utils.py
```diff
@@ -356,3 +361,37 @@ def get_local_path(source, destination=None, create_temp=False):
 def explicit_tenant_name_message(tenant_name, logger):
     if tenant_name:
         logger.info('Explicitly using tenant `{0}`'.format(tenant_name))
+
+
+def validate_date(date):
```
inbala: Why "validate_date" and not "parse_datetime" as before? (because it is not just a validation check)

PR #904, opened 2019-01-08
deployment-update-preview

yarivgiga:


cloudify_cli/cli/helptexts.py
```diff
@@ -88,6 +88,11 @@
     "In deployment update, perform install workflow and then uninstall "
     "workflow. default: uninstall and then install"
 )
+PREVIEW = (
+    "In deployment update, perform a preview of an update and not an actual "
+    "update. The changes that should be done will be returned, but no change "
+    "will actually take effect"
```
geokala: This doesn't need to state that it's 'in deployment update'. If we feel that we need to track that then we should call this PREVIEW_UPDATE or similar instead. The user will see this in the helptext for the deployment update, so stating it in the message itself is redundant.

Also, the message might be better formatted something like:
"Preview the deployment update, stating what changes will be made without actually applying any changes." (or something similar)

cloudify_cli/commands/deployments.py
```diff
@@ -54,10 +59,10 @@
 DEPLOYMENT_UPDATE_COLUMNS = ['id', 'deployment_id', 'tenant_name', 'state',
                              'execution_id', 'created_at', 'visibility',
                              'old_blueprint_id', 'new_blueprint_id']
+NON_PREVIEW_COLUMNS = ['id', 'execution_id']
+STEPS_COLUMNS = ['entity_type', 'entity_id', 'action']
 TENANT_HELP_MESSAGE = 'The name of the tenant of the deployment'
-DEPLOYMENTS_SUMMARY_FIELDS = [
-    'blueprint_id',
-] + BASE_SUMMARY_FIELDS
+DEPLOYMENTS_SUMMARY_FIELDS = ['blueprint_id'] + BASE_SUMMARY_FIELDS
```
geokala: Why is this modified in an unrelated commit?

cloudify_cli/commands/deployments.py
```diff
@@ -68,6 +73,67 @@ def deployments():
...
+                         explicit_reinstall=None,
+                         preview=False,
+                         skip_install=False,
+                         skip_uninstall=False,
+                         skip_reinstall=False):
+    if explicit_reinstall is None:
+        explicit_reinstall = []
+    columns = DEPLOYMENT_UPDATE_COLUMNS
+    if preview:
+        columns = [c for c in columns if c not in NON_PREVIEW_COLUMNS]
```
geokala: This being a double negative isn't as nice as it could be- maybe we could reverse the column listing above to make it a set with some elements removed and then we wouldn't need to do this?

cloudify_cli/commands/deployments.py
```diff
@@ -68,6 +73,67 @@ def deployments():
...
+            continue
+        if step['action'] == 'add':
+            deployment_update_dict['installed_nodes'].append(entity[1])
+        elif step['action'] == 'remove':
+            deployment_update_dict['uninstalled_nodes'].append(entity[1])
+        elif step['action'] == 'modify':
+            deployment_update_dict['reinstalled_nodes'].append(entity[1])
+
+    if get_global_json_output():
+        columns += [
```
geokala: We store all other column listings as constants- why not these ones?

cloudify_cli/commands/deployments.py
```diff
@@ -312,6 +368,7 @@ def manager_update(ctx,
     if blueprint_id:
         msg += ', using blueprint {0}'.format(blueprint_id)
     logger.info(msg)
+    reinstall_list = list(reinstall_list or [])
```
geokala: Would this be better as list(reinstall_list) if reinstall_list else []
(to avoid the extra cast to list)

cloudify_cli/table.py
```diff
@@ -136,13 +143,13 @@ def print_details(data, title):
...
     if get_global_json_output():
         output(json.dumps(data, cls=CloudifyJSONEncoder))
-    else:
-        output(title)
-        for item in data.items():
-            field_name = str(item[0]) + ':'
-            field_value = str(item[1])
-            field_value = get_timestamp(field_value) or field_value
-            output('\t{0} {1}'.format(field_name.ljust(16), field_value))
+        return
```
geokala: Any reason we prefer changing this to use a return instead of an else? I think the else reads more nicely, and this appears to otherwise be a bunch of changes that are unrelated to the actual change?

cloudify_cli/tests/commands/test_deployments.py
```diff
@@ -64,22 +64,32 @@ def setUp(self):
     def test_deployment_update_get(self):
```
geokala: We don't appear to have a test that confirms that a deployment update with preview actually leaves the deployment alone?

cloudify_cli/commands/deployments.py
tehasdf: no need for the `list()` call at all, actually, it's already a list

tehasdf:
@geokala updated, please take a look again

PR #915, opened 2019-02-17
CY-1056: fixed multiple issues with agents install/validate

isaac-s:
Fixed various issues with filtration logic and workflow execution logic.
The intention is to schedule the `install_new_agents` execution only for deployments that actually have node instances which match the filters.

Also, I removed the functionality of tracking the execution logs because it is impossible to interrupt using `SIGINT` (due to the use of threads). Instead, we tell the user which executions where scheduled for each affected deployment.

### DIFF
--- a/cloudify_cli/commands/agents.py
+++ b/cloudify_cli/commands/agents.py
@@ -40,19 +40,6 @@ AGENT_COLUMNS = ['id', 'ip', 'deployment', 'node', 'system', 'version',
 MAX_TRACKER_THREADS = 20
 
 
-def _handle_deployment_id(logger, deployment, agent_filters):
-    # Handle the case when a deployment ID is provided as a positional
-    # argument.
-    if deployment:
-        logger.warning('Passing the deployment ID as an argument is '
-                       'deprecated, use --deployment-id instead')
-        if agent_filters[cfy.AGENT_FILTER_DEPLOYMENT_ID]:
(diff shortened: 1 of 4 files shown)


cloudify_cli/commands/agents.py
```diff
@@ -44,23 +57,6 @@ def agents():
     pass
 
 
-def _is_deployment_installed(client, deployment_id):
```
isaac-s: This used to test that all node instances are in `started` state.
In fact, only `Compute` instances need to have this check done.
The corrected test has been moved to the manager side (see https://github.com/cloudify-cosmo/cloudify-common/pull/216)

cloudify_cli/commands/agents.py
```diff
@@ -120,185 +103,146 @@ def install(deployment,
...
+                deployment_id, workflow_id, execution_params,
                 allow_custom_parameters=True)
-            execution = wait_for_execution(
-                client,
-                execution,
-                events_handler=threadsafe_events_logger,
-                include_logs=include_logs,
-                timeout=timeout
+            logger.info(
+                "Scheduled execution for deployment '%s': %s",
```
idob-cloudify: why "scheduled" and not "started"?

isaac-s: @idob-cloudify committed a change for that.

PR #955, opened 2019-05-29
Execution id parameter: allow workflow id

tehasdf:
Instead of just working with the execution id, it is useful to
provide a shorthand to use the last created execution for a given
workflow.

With this, the following calls are now allowed:
    cfy exec cancel install
    cfy exec resume install
    cfy events list uninstall
    cfy exec get execute_operation
etc

### DIFF
--- a/cloudify_cli/cli/cfy.py
+++ b/cloudify_cli/cli/cfy.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 ############
 
-
+import re
 import sys
 import urllib
 import difflib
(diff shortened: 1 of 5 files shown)


cloudify_cli/cli/cfy.py
```diff
@@ -155,6 +159,29 @@ def inputs_callback(ctx, param, value):
     return inputs_to_dict(value)
 
 
+def _get_execution_id(ctx, param, value):
+    tenant = ctx.params.get('tenant_name')
+    is_uuid = value and re.match('[a-fA-F0-9-]{30,40}', value)
```
geokala: Wouldn't it be better to try to select it from the execution IDs?

tehasdf: don't really want to add an additional request

geokala:

PR #963, opened 2019-06-13
CY-1060 Add plugins update commands

davidginzbourg:
- Adds a flag the dont_update_plugins flag to deployment updates

### DIFF
--- a/cloudify_cli/commands/plugins.py
+++ b/cloudify_cli/commands/plugins.py
@@ -18,6 +18,10 @@ import os
 
 import wagon
 
+from cloudify_cli import execution_events_fetcher
+from cloudify_cli.logger import get_events_logger
+from cloudify_cli.exceptions import SuppressedCloudifyCliError
+
 from cloudify_rest_client.constants import VISIBILITY_EXCEPT_PRIVATE
 
 from .. import utils
@@ -30,6 +34,9 @@ from ..utils import (prettify_client_error,
 PLUGIN_COLUMNS = ['id', 'package_name', 'package_version', 'distribution',
                   'supported_platform', 'distribution_release', 'uploaded_at',
                   'visibility', 'tenant_name', 'created_by', 'yaml_url_path']
+PLUGINS_UPDATE_COLUMNS = ['id', 'state', 'blueprint_id', 'temp_blueprint_id',
+                          'execution_id', 'deployments_to_update',
+                          'visibility', 'created_at', 'forced']
 GET_DATA_COLUMNS = ['file_server_path']
 EXCLUDED_COLUMNS = ['archive_name', 'distribution_version', 'excluded_wheels',
                     'package_source', 'supported_py_versions', 'wheels']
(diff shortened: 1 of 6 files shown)


cloudify_cli/commands/plugins.py
```diff
@@ -30,6 +34,9 @@
 PLUGIN_COLUMNS = ['id', 'package_name', 'package_version', 'distribution',
                   'supported_platform', 'distribution_release', 'uploaded_at',
                   'visibility', 'tenant_name', 'created_by', 'yaml_url_path']
+PLUGINS_UPDATE_COLUMNS = ['id', 'state', 'blueprint_id', 'temp_blueprint_id',
```
tyacbovi: is there more col that are not here?

cloudify_cli/commands/plugins.py
```diff
@@ -285,3 +292,135 @@ def set_visibility(plugin_id, visibility, logger, client):
...
+    logger.info('Updating the plugins of the deployments of the blueprint '
+                '{}'.format(blueprint_id))
+    plugins_update = client.plugins_update.update_plugins(blueprint_id, force)
+    events_logger = get_events_logger(json_output)
+    execution = execution_events_fetcher.wait_for_execution(
+        client,
+        client.executions.get(plugins_update.execution_id),
+        events_handler=events_logger,
+        include_logs=include_logs,
+        timeout=None  # don't timeout ever
```
tyacbovi: Sure about this? maybe set a long timeout and give the user the option to change it

cloudify_cli/tests/commands/test_plugins.py
```diff
@@ -116,3 +120,109 @@ def test_plugins_upload_with_visibility(self):
...
+        self.client.plugins_update = MagicMock()
+        self._mock_wait_for_executions(False)
+
+    def test_plugins_get(self):
+        self.client.plugins_update.get = MagicMock(
+            return_value=plugins_update.PluginsUpdate({
+                'id': 'asdf'
+            }))
+        outcome = self.invoke('cfy plugins get-update asdf')
+        # One time in the "Retrieving..." log message and one time in the
```
tyacbovi: can be deleted

cloudify_cli/tests/commands/test_plugins.py
```diff
@@ -116,3 +120,109 @@ def test_plugins_upload_with_visibility(self):
...
+        patcher.start()
+
+    def setUp(self):
+        super(PluginsUpdateTest, self).setUp()
+        self.use_manager()
+        self.client.executions = MagicMock()
+        self.client.plugins_update = MagicMock()
+        self._mock_wait_for_executions(False)
+
+    def test_plugins_get(self):
```
tyacbovi: what about getting not valid id

cloudify_cli/commands/plugins.py
davidginzbourg: Nope, that's about it

cloudify_cli/commands/plugins.py
davidginzbourg: I don't think we should trust the user to give the right timeout because the timeout is dependant on the amount of active deployments there are for that blueprint

davidginzbourg: I think system workflows don't have a timeout regardless

cloudify_cli/tests/commands/test_plugins.py
davidginzbourg: That should be the click option responsibility since the IDs here have the same rules for all resourced.

tyacbovi: Not sure about that

davidginzbourg: https://github.com/cloudify-cosmo/cloudify-cli/blob/master/cloudify_cli/cli/cfy.py there's a validation function there. I think I should add the plugins_update_id arg option to the options there. Same for the deployment update.

tyacbovi: Test what happen if get not existing dep id

PR #965, opened 2019-06-19
Print the runtime properties after update

inbala:


cloudify_cli/commands/node_instances.py
```diff
@@ -249,8 +234,10 @@ def _modify_runtime(node_instance_id, logger, client, tenant_name,
     client.node_instances.update(node_instance_id,
                                  runtime_properties=runtime_properties,
                                  version=new_version)
-    logger.info('Updated the runtime properties of node instance '
-                '{0}'.format(node_instance_id))
+    logger.info('Successfully updated the runtime properties of {0}'
```
tyacbovi:
```suggestion
    logger.info('Successfully updated the runtime properties of "{0}"'
```
it's more clear

PR #979, opened 2019-08-01
CY-1538 New 'cfy snapshots status' CLI Command

adigrabow:


cloudify_cli/commands/snapshots.py
```diff
@@ -57,16 +57,16 @@ def restore(snapshot_id,
     """
     logger.info('Restoring snapshot {0}...'.format(snapshot_id))
     recreate_deployments_envs = not without_deployment_envs
-    execution = client.snapshots.restore(
```
tyacbovi: You can leave the msg, more info is not bad :)

PR #988, opened 2019-09-02
CY-1507 Deployments: add the "runtime only evaluation" flags

tehasdf:


tehasdf:
this was merged into the UI and other parts months ago, but this PR was missed then
(integration tests use rest-client, and maunal tests mostly use the UI, so we missed the CLI)

PR #991, opened 2019-09-04
CY-1656 Fix license file in Windows cloudify-cli installer

glukhman:


packaging/LICENSE
```diff
@@ -1,32 +1,213 @@
...
+                           Cloudify Software License Agreement
```
tyacbovi: do we use this one?

PR #998, opened 2019-09-23
CY-1598 Update the status command

inbala:


cloudify_cli/commands/status.py
```diff
@@ -55,13 +55,10 @@ def status(logger, client):
...
-    for service in status_result['services']:
-        state = service['instances'][0]['state'] \
-            if 'instances' in service and \
-               len(service['instances']) > 0 else 'unknown'
+    for display_name, service in status_result['services'].items():
         services.append({
-            'service': service['display_name'].ljust(30),
-            'status': state
+            'service': display_name.ljust(30),
+            'status': service['status']
```
tyacbovi: So we don't show here the extra data?

inbala: No, I wanted to keep it as it was.
But I can ask Ofer.

PR #1015, opened 2019-11-26
CY-1753 Naive remove diamond from Win CLI

glukhman:


cloudify_cli/commands/init.py
```diff
@@ -200,6 +200,16 @@ def set_config(enable_colors=False):
         log_path=DEFAULT_LOG_FILE,
         enable_colors=enable_colors
     )
+    # for windows
+    win_config_raw = None
+    win_config_path = os.path.join(env.CLOUDIFY_WORKDIR, 'win_config.yaml')
```
tyacbovi: I think ypu should put this into a new func and add some doc around this issue

PR #1019, opened 2019-12-10
CY-2066: status reports change to the new place

tyacbovi:


cloudify_cli/commands/logs.py
```diff
@@ -53,7 +53,8 @@ def _archive_logs(logger,
     # knowingly. We don't want to run another ssh command just to verify
     # something that will almost never happen.
     ssh.run_command_on_manager(
-        'tar -czf {0} -C /var/log cloudify -C /opt/cloudify cluster_statuses '
+        'tar -czf {0} -C /var/log cloudify -C '
```
davidginzbourg: Please put the -C at the beginning of the next row

PR #1027, opened 2019-12-31
CY-2153 Define default_timeout_sec in init

inbala:


cloudify_cli/env.py
```diff
@@ -529,6 +528,7 @@ def __init__(self, *args, **kwargs):
         first_node = self._cluster[0]
         self.cert = first_node.get('cert') or self.cert
         self.trust_all = first_node.get('trust_all') or self.trust_all
+        self.default_timeout_sec = (5, None)
```
tyacbovi: Why not pass it to the super's init?

PR #1029, opened 2020-01-01
CY-2156 change profile.cluster attribute to dict

YoniItzhak:


cloudify_cli/commands/cluster.py
```diff
@@ -198,28 +217,48 @@ def _update_profile_cluster_settings(nodes, logger=None):
...
-        env.profile.cluster = []
+        env.profile.cluster = dict()
+    _update_cluster_nodes(manager_nodes, CloudifyNodeType.MANAGER, logger)
+    _update_cluster_nodes(broker_nodes, CloudifyNodeType.BROKER, logger)
+    _update_cluster_nodes(db_nodes, CloudifyNodeType.DB, logger)
+
+
+def _update_cluster_nodes(nodes, nodes_type, logger):
+    stored_nodes = env.profile.cluster.get(nodes_type)
+    stored_nodes_names = ({node['hostname'] for node in stored_nodes}
```
tyacbovi: 'hostname' let's move it to const

cloudify_cli/commands/cluster.py
```diff
@@ -198,28 +217,48 @@ def _update_profile_cluster_settings(nodes, logger=None):
...
     # filter out removed nodes
-    env.profile.cluster = [n for n in env.profile.cluster
-                           if n['hostname'] in received_nodes]
+    env.profile.cluster[nodes_type] = [
+        node for node in env.profile.cluster[nodes_type]
+        if (node.get('hostname') or node.get('name')) in received_nodes_names]
     env.profile.save()
 
 
+def _update_node(node, nodes_type, logger, stored_nodes_names):
```
tyacbovi: nodes_type => node_type

cloudify_cli/commands/cluster.py
```diff
@@ -198,28 +217,48 @@ def _update_profile_cluster_settings(nodes, logger=None):
...
-                # ones from the last used manager
-                'manager_ip': node_ip
-            })
+        _update_node(node, nodes_type, logger, stored_nodes_names)
     # filter out removed nodes
-    env.profile.cluster = [n for n in env.profile.cluster
-                           if n['hostname'] in received_nodes]
+    env.profile.cluster[nodes_type] = [
+        node for node in env.profile.cluster[nodes_type]
+        if (node.get('hostname') or node.get('name')) in received_nodes_names]
```
tyacbovi: node.get('hostname') or node.get('name') this repeats a lot

cloudify_cli/commands/cluster.py
```diff
@@ -198,28 +217,48 @@ def _update_profile_cluster_settings(nodes, logger=None):
     received nodes, because the profile might have more details about
     the nodes (eg. a certificate path)
     """
-    stored_nodes = {node['hostname'] for node in env.profile.cluster}
-    received_nodes = {node['hostname'] for node in nodes}
     if env.profile.cluster is None:
```
tyacbovi: env.profile.cluster = env.profile.cluster or dict()

cloudify_cli/env.py
```diff
@@ -32,6 +32,7 @@
 from cloudify_rest_client import CloudifyClient
 from cloudify_rest_client.client import HTTPClient
 from cloudify_rest_client.exceptions import CloudifyClientError
+from cloudify.cluster_status import CloudifyNodeType
```
tyacbovi: order by len and add an empty line space

cloudify_cli/commands/cluster.py
```diff
@@ -198,28 +218,50 @@ def _update_profile_cluster_settings(nodes, logger=None):
...
-        env.profile.cluster = []
+    env.profile.cluster = env.profile.cluster or dict()
+    _update_cluster_nodes(manager_nodes, CloudifyNodeType.MANAGER, logger)
+    _update_cluster_nodes(broker_nodes, CloudifyNodeType.BROKER, logger)
+    _update_cluster_nodes(db_nodes, CloudifyNodeType.DB, logger)
+
+
+def _update_cluster_nodes(nodes, nodes_type, logger):
+    stored_nodes = env.profile.cluster.get(nodes_type)
+    stored_nodes_names = ({_get_node_name(node) for node in stored_nodes}
```
tyacbovi: well more like _get_node_name=> _get_node_host

cloudify_cli/commands/profiles.py
```diff
@@ -33,6 +33,8 @@
 from ..cli import helptexts
 from ..exceptions import CloudifyCliError
 from ..table import print_data, print_single
+from ..env import get_rest_client
```
tyacbovi: order by len

cloudify_cli/commands/profiles.py
```diff
@@ -189,13 +190,20 @@ def use(manager_ip,
...
-    if cluster:
-        client = env.get_rest_client(
-            client_profile=env.profile,
-            username=(env.profile.manager_username or 'admin'),
-            password=(env.profile.manager_password or 'admin')
-        )
+
+    active_profile = _get_profile(env.get_active_profile())
+    active_profile_context = env.get_profile_context(env.get_active_profile())
+    if active_profile_context.manager_username is None:     # docl 1st run
```
tyacbovi: what do u mean here?

cloudify_cli/commands/profiles.py
```diff
@@ -189,13 +190,20 @@ def use(manager_ip,
...
-    cluster = env.profile.cluster.get(CloudifyNodeType.MANAGER)
-    if cluster:
-        client = env.get_rest_client(
-            client_profile=env.profile,
-            username=(env.profile.manager_username or 'admin'),
-            password=(env.profile.manager_password or 'admin')
-        )
+
+    active_profile = _get_profile(env.get_active_profile())
+    active_profile_context = env.get_profile_context(env.get_active_profile())
```
tyacbovi: you can run this once 'env.get_active_profile()'

cloudify_cli/commands/cluster.py
```diff
@@ -100,8 +100,14 @@ def _inner(client, *args, **kwargs):
 
 def _all_in_one_manager(client):
     manager_nodes = client.manager.get_managers().items
+    if len(manager_nodes) > 1:
```
tyacbovi: this is not correct check, all of them need to be one

cloudify_cli/commands/profiles.py
YoniItzhak: Just using `get_profile_context` should work. If no profile name is transferred to this function, it will return the current profile context. This comment relates to the next one.

cloudify_cli/commands/profiles.py
```diff
@@ -187,6 +191,21 @@ def use(manager_ip,
             logger=logger,
             **kwargs)
 
+    active_profile = _get_profile(env.get_active_profile())
+    active_profile_context = env.get_profile_context(env.get_active_profile())
+    if active_profile_context.manager_username is None:     # docl 1st run
+        return
+    client = get_rest_client(
```
YoniItzhak: Using the `get_rest_client` would use the current profile, in case no client_profilem is given to the function. Moreover `get_rest_client` resolves the `rest_port`, `rest_protocol`, `rest_certificate`, and `tenant_name` the same as you've done here.

cloudify_cli/commands/profiles.py
```diff
@@ -187,6 +191,21 @@ def use(manager_ip,
...
+        rest_port=active_profile['rest_port'],
+        rest_protocol=active_profile['rest_protocol'],
+        rest_cert=active_profile['rest_certificate'],
+        username=active_profile_context.manager_username or 'admin',
```
YoniItzhak: In which case the 'active_profile_context.manager_username' would be None? The same goes for the 'manager_password'. According to the command `cfy profiles use,` the username and password are mandatory.

cloudify_cli/commands/cluster.py
```diff
@@ -89,15 +90,37 @@ def _deco(f):
...
+def _all_in_one_manager(client):
+    manager_nodes = client.manager.get_managers().items
+    if len(manager_nodes) > 1:
+        return False
+    broker_nodes = client.manager.get_brokers().items
+    if len(broker_nodes) > 1:
+        return False
+    db_nodes = client.manager.get_db_nodes().items
+    if len(db_nodes) > 1:
+        return False
```
YoniItzhak: Why are the added lines necessary for the assertion?

cloudify_cli/commands/profiles.py
YoniItzhak: You can write the same as `if **not** active_profile_context.manager_username:`

cloudify_cli/commands/profiles.py
YoniItzhak: And a better practice would be using the `get` function in these cases, i.e. `active_profile[x]` => `active_profile.get(x)`. 
In that way we avoid key errors, in case there are ones.

cloudify_cli/tests/commands/test_profiles.py
```diff
@@ -371,9 +373,11 @@ def test_use_cannot_update_profile(self, *_):
         self.assertIn('The passed in options are ignored: manager_password',
                       outcome.logs)
 
+    @patch('cloudify_cli.commands.profiles._all_in_one_manager',
+           return_value=True)
     @patch('cloudify_cli.commands.profiles._get_provider_context',
            return_value={})
-    def test_use_existing_only_switches(self, mock_get_context):
+    def test_use_existing_only_switches(self, mock_get_context, _):
```
YoniItzhak: Why is `_` passed to the function?

cloudify_cli/commands/profiles.py
glukhman: This function runs automatically as part of _docl run_, in which case there's no username nor password

cloudify_cli/commands/cluster.py
glukhman: I figured out short-circuiting the check with as few rest calls as possible would be a good thing

cloudify_cli/tests/commands/test_profiles.py
glukhman: for the additional patch

cloudify_cli/commands/cluster.py
glukhman: but it suffices that we have more than 1 of any of the node types to determine it's not an AIO

cloudify_cli/commands/profiles.py
glukhman: This function runs automatically as part of docl run, in which case there's no username nor password. L196 is for that case

cloudify_cli/commands/cluster.py
tyacbovi: Well, there is not true, because you can have a cluster with one manager  also for the other but not together

cloudify_cli/commands/profiles.py
```diff
@@ -190,19 +190,12 @@ def use(manager_ip,
             skip_credentials_validation=skip_credentials_validation,
             logger=logger,
             **kwargs)
-
-    active_profile = _get_profile(env.get_active_profile())
-    active_profile_context = env.get_profile_context(env.get_active_profile())
-    if active_profile_context.manager_username is None:     # docl 1st run
+    _get_profile(env.get_active_profile())
+    active_profile_context = env.get_profile_context()
+    if not active_profile_context.manager_username:     # docl 1st run
```
tyacbovi: please remove the docl comment it's not really relevant in the code

cloudify_cli/commands/profiles.py
```diff
@@ -192,6 +192,8 @@ def use(manager_ip,
             **kwargs)
     if not env.profile.manager_username:
         return
+    if type(env.profile.cluster) == type([]):   # noqa
```
YoniItzhak: isinstance(env.profile.cluster, list)

cloudify_cli/commands/profiles.py
```diff
@@ -256,6 +258,8 @@ def _create_profile(
         skip_credentials_validation
     )
     init.init_manager_profile(profile_name=profile_name)
+    env.set_active_profile(profile_name)
```
YoniItzhak: It also shows in line 280

PR #1033, opened 2020-01-05
CY-2161 update cfy profiles command

YoniItzhak:


cloudify_cli/commands/profiles.py
```diff
@@ -389,9 +390,9 @@ def _set_profile_ssl(ssl, rest_port, logger):
     env.profile.rest_port = port
     env.profile.rest_protocol = protocol
 
-    if env.profile.cluster:
+    if env.profile.cluster.get(CloudifyNodeType.MANAGER):
```
tyacbovi: You can save the managers list aside

PR #1035, opened 2020-01-08
CY-2163 updating `cfy logs` command

YoniItzhak:


cloudify_cli/commands/logs.py
```diff
@@ -32,35 +32,37 @@ def logs():
...
     logger.info('Creating logs archive in manager: {0}'.format(archive_path))
     # We skip checking if the tar executable can be found on the machine
     # knowingly. We don't want to run another ssh command just to verify
     # something that will almost never happen.
-    ssh.run_command_on_manager(
+    ssh.run_command_on_host(
         'tar -czf {0} -C /var/log cloudify '
         '-C /opt/manager cluster_statuses '
-        '--ignore-failed-read'.format(archive_path),
+        '--warning=no-file-changed'.format(archive_path),
```
tyacbovi: Why this change was needed?

cloudify_cli/commands/logs.py
```diff
@@ -79,42 +81,54 @@ def download(output_path, all_nodes, logger):
...
             raise CloudifyCliError(
                 "No cluster nodes defined in this profile")
-        for node in env.profile.cluster:
+        cluster_nodes = _get_cluster_nodes()
+        for node in cluster_nodes:
             ssh_user = node.get('ssh_user') or env.profile.ssh_user
             ssh_key = node.get('ssh_key') or env.profile.ssh_key
             if not ssh_user or not ssh_key:
-                logger.info('No ssh details defined for manager {0} in '
+                logger.info('No ssh details defined for host {0} in '
```
tyacbovi: and maybe add host_string

cloudify_cli/commands/logs.py
```diff
@@ -79,42 +81,54 @@ def download(output_path, all_nodes, logger):
         if not env.profile.cluster:
```
tyacbovi: is this cond still good enough ?

cloudify_cli/commands/logs.py
```diff
@@ -79,42 +81,54 @@ def download(output_path, all_nodes, logger):
...
-        ssh.run_command_on_manager(
-            'rm {0}'.format(archive_path_on_manager), use_sudo=True)
+        ssh.run_command_on_host('rm {0}'.format(archive_path_on_manager),
+                                host_string,
+                                use_sudo=True)
+
+
+def _get_cluster_nodes():
+    cluster_nodes = []
+    for nodes_list in env.profile.cluster.values():
```
tyacbovi: Why not just add all list and not go over one by one?

cloudify_cli/commands/logs.py
```diff
@@ -79,42 +81,54 @@ def download(output_path, all_nodes, logger):
...
+                                   host_string=host_string,
+                                   key_filename=ssh_key)
+            logger.info('Removing archive from host...')
+            ssh.run_command_on_host('rm {0}'.format(archive_path_on_host),
+                                    use_sudo=True,
+                                    host_string=host_string,
+                                    key_filename=ssh_key)
     else:
-        archive_path_on_manager = _archive_logs(logger)
+        host_string = env.build_manager_host_string()
```
tyacbovi: host_string is a mandatory input, so why override it?

cloudify_cli/commands/logs.py
YoniItzhak: According to [man tar](http://man7.org/linux/man-pages/man1/tar.1.html), `--warning=no-file-changed` is used to cover the case where the file is changed while it is being read, not `--ignore-failed-read`. The only thing is that it still returns 1 as the return code, meaning it failed. Because of that, I added the following command to verify the tar was truly created. 
As a remark, while trying to run `cfy logs download` on a cluster manager node before changing the flag (i.e. while it was --ignore-failed-read) I got an error of file is changed while it is being read.

cloudify_cli/commands/logs.py
YoniItzhak: Yes, since if it is an AIO Manager, the cluster attribute would be an empty dictionary.

cloudify_cli/commands/logs.py
YoniItzhak: Great point

cloudify_cli/commands/logs.py
YoniItzhak: What do you mean by overriding it?

tyacbovi: The input to this function contains 'host_string'

PR #1038, opened 2020-01-16
CY-2212: allowing to run cluster related commands on AIO

tyacbovi:
but with a warning to the user, because AIO is still a kind of a cluster


cloudify_cli/commands/cluster.py
```diff
@@ -94,8 +94,9 @@ def _deco(f):
         @wraps(f)
         def _inner(client, *args, **kwargs):
             if _all_in_one_manager(client):
-                raise CloudifyCliError('You cannot run cluster actions on an '
-                                       'all-in-one Manager')
+                get_logger().warning('Notice: You are trying to run cluster '
```
geokala: Should this perhaps say "Warning: " instead of "Notice: "?

PR #1043, opened 2020-01-21
CY-2227 When in cluster mode, for every rest request first use the ma…

inbala:
…nager given in profiles use

### DIFF
--- a/cloudify_cli/env.py
+++ b/cloudify_cli/env.py
@@ -548,21 +548,35 @@ class ClusterHTTPClient(HTTPClient):
         if kwargs.get('timeout') is None:
             kwargs['timeout'] = self.default_timeout_sec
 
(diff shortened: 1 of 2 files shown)


cloudify_cli/env.py
```diff
@@ -548,22 +548,34 @@ def do_request(self, *args, **kwargs):
...
             kwargs['timeout'] = self.default_timeout_sec
 
+        # First try with the main manager ip given when creating the profile
+        # with `cfy profiles use`
+        self.host = self._profile.manager_ip
+        response = self._try_do_request(*args, **kwargs)
+        if response:
+            return response
+
         for node_index, node in list(enumerate(
```
tyacbovi: This will repeat the first server twice, no?
And how did this behavior changed?

PR #1044, opened 2020-01-22
CY-2249 `ssl enable`: don't update-profile with cluster nodes

tehasdf:
`cfy ssl enable` is supposed to just update the profile to enable ssl

It attempted to also do the equivalent of `cfy cluster update-profile`,
however that requires to contact the manager, and that isn't
necessarily possible (yet) because the manager might not have become
ssl (yet).

Anyway, we're updating the ssl setting for each manager. We're just
not updating the profile list (with any new managers that might
have been added in the meantime)

Alternatively we could've added a sleep, but just Not Doing It™
is simpler

### DIFF
--- a/cloudify_cli/commands/profiles.py
+++ b/cloudify_cli/commands/profiles.py
@@ -402,7 +402,6 @@ def _set_profile_ssl(ssl, rest_port, logger):
     env.profile.rest_port = port
     env.profile.rest_protocol = protocol
 
-    _update_cluster_profile_to_dict(logger)
     manager_cluster = env.profile.cluster.get(CloudifyNodeType.MANAGER)
     if manager_cluster:
         missing_certs = []


tehasdf:
This is mostly to fix the test_ssl systest (CY-2249)

@YoniItzhak note this; perhaps in the future we'll want to replace this with a sleep or something

Ideally, we can just lose `cfy ssl` entirely, because it's pretty stupid imo.

geokala:
Approved and yes, let's just get rid of ssl as an option and make it be 'the way' (TM).

YoniItzhak:
Thanks for the update @tehasdf. I added this line to the code to support the following use-case:
1. The user has an old profile saved on his CLI client. "old" means profile.cluster is a list (not a dictionary).
2. The user updates the CLI, and runs `cfy profiles set --ssl <on/off>`.
3. The code would go to this function [_set_profile_ssl](https://github.com/cloudify-cosmo/cloudify-cli/blob/87f1deeb5944cd91b4321da185123f16c49bc85d/cloudify_cli/commands/profiles.py#L389)
and here would try to run [cluster.get](https://github.com/cloudify-cosmo/cloudify-cli/blob/87f1deeb5944cd91b4321da185123f16c49bc85d/cloudify_cli/commands/profiles.py#L405).

This use case is based on [Isaak's comment](https://cloudify.slack.com/archives/G2D9N51LP/p1579381917000900)

Did you check this use case? Or is there an alternative solution? 

Thanks again

PR #1045, opened 2020-01-23
CY-2233 Add an option to connect to a specific manager by IP or host

inbala:


cloudify_cli/cli/helptexts.py
```diff
@@ -388,3 +388,4 @@
                           "evaluated at runtime, and no intrinsic functions "\
                           "will be evaluated at parse time (such as "\
                           "get_input, get_property)"
+MANAGER = "Connect to a specific manager by IP or host"
```
tyacbovi: Will allow directing the command to ...

Or in line of this?

cloudify_cli/cli/cfy.py
```diff
@@ -231,6 +231,18 @@ def set_format(ctx, param, value):
     return value
 
 
+def set_manager(ctx, param, value):
+    if value is not None:
```
tyacbovi: When this will be None?
And maybe we will want to log that

PR #1061, opened 2020-03-09
Get current version based on package version

tehasdf:
Drop the VERSION file because it doesn't make sense to keep
the version number in 2 places.
Also, the 'edition' wasn't even used.

### DIFF
--- a/cloudify_cli/VERSION
+++ b/cloudify_cli/VERSION
@@ -1,4 +1,0 @@
-{
-  "edition": "premium",
-  "version": "5.1.0-.dev1"
-}
--- a/cloudify_cli/cli/cfy.py
+++ b/cloudify_cli/cli/cfy.py
@@ -21,6 +21,7 @@ import difflib
 import StringIO
 import warnings
 import traceback
+import pkg_resources
 from functools import wraps
 
 import click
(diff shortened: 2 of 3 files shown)


cloudify_cli/VERSION
```diff
@@ -1,4 +0,0 @@
-{
-  "edition": "premium",
-  "version": "5.1.0-.dev1"
-}
```
mabuaisha: This will affect the release process for both community and premium,

tehasdf: @mabuaisha where?

mabuaisha: In the build-system but I think we should be fine since there is no something specific to cli defined inside the config.yaml https://github.com/cloudify-cosmo/cloudify-build-system/blob/master/config/core-config.yaml#L82

And I think for community we are bump the community version inside `setup.py` which I think will be fine https://github.com/cloudify-cosmo/cloudify-cli/blob/20.03.03-build/setup.py#L21

tehasdf: yes, good, so this doesn't affect anything after all

and bumping this file can be removed as well of course

tehasdf: or bumping doesn't need to be removed (yet) because i think we still have a VERSION file somewhere? (rest?)

mabuaisha: well, it search for this regex https://github.com/cloudify-cosmo/cloudify-build-system/blob/master/config/core-config.yaml#L26 which means if this file does not exist on the cli then it will try to match on agent and I think there is another one on cloudify-manager-install 

I think this should be fine because at the end of the day this file used in order to grap the version of the current cli and since you depend on python to get that version then we are fine I think

PR #1063, opened 2020-03-11
CY-2362 Python 3 porting: stage 1

tehasdf:
Simple stuff, futurize --stage1


.circleci/config.yml
```diff
@@ -64,13 +64,41 @@ jobs:
...
+      - run:
+          name: install futurize
+          command: pip install future --user
+      - run:
+          name: find python3-incompatible code
+          command: |
+            FUTURIZE="futurize ."
+            while read line; do
+              [[ "$line" =~ ^#.* ]] && continue
+              FUTURIZE="${FUTURIZE} ${line}"
```
tehasdf: note that this is different from the cloudify-common by this line, because in common we do `-f ${line}`, and here just `${line}` which means that the file contains flags, eg. `--stage1` or `-f specific.fixer`.

This was more useful here, because there's less changes, so it's easier to spell out `--stage1` than the 30 individual fixers that go into it.

This is probably a better way, and we should do that everywhere

cloudify_cli/commands/profiles.py
```diff
@@ -114,7 +114,7 @@ def show(logger):
                   short_help='List profiles')
 @cfy.options.common_options
 @cfy.pass_logger
-def list(logger):
+def profiles_list(logger):
```
tehasdf: renaming this to stop shadowing the builtin `list` name

cloudify_cli/commands/secrets.py
```diff
@@ -355,8 +356,8 @@ def _print_secrets_errors(secrets_errors_dict, logger):
                 ' to the the errors mentioned for each secret. The secrets`'
                 ' number refer to their position in the imported list:')
     for key, secret_errors in secrets_errors_list:
-        print('\n\tSecret {0}:'.format(int(key) + 1))
+        click.echo('\n\tSecret {0}:'.format(int(key) + 1))
```
tehasdf: applying the print fixer by getting rid of prints, yay

PR #1068, opened 2020-03-12
Make `cfy ssh` and fabric optional

tehasdf:
For now, make the ssh subsystem an optional dependency.
This is so that the CLI can work with the current
cloudify-fabric-plugin.

This can/will possibly be reverted later, after the fabric-plugin
is ported, or after plugins are made to be installed
into separate virtualenvs.

For now, this should fix the build, anyway.

### DIFF
--- a/cloudify_cli/commands/ssh.py
+++ b/cloudify_cli/commands/ssh.py
@@ -16,12 +16,16 @@
 
 import re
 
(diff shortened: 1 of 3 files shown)


tehasdf:
CY-2457 is about updating fabric version in cloudify-fabric-plugin

PR #1071, opened 2020-03-17
CY-2436 Add a .spec for building a CLI rpm

tehasdf:


mabuaisha:
Unless we want to keep the job in jenkins for centos, then we may keep that DockerFile

PR #1072, opened 2020-03-17
CY-2436 add a .deb build

tehasdf:


.circleci/config.yml
```diff
@@ -84,7 +84,40 @@ jobs:
...
+      - image: debian:9
+    steps:
+      - checkout
+      - run:
+          name: Set envvars
+          command: |
+            echo "export PROJECT_DIR=~/project" >>$BASH_ENV
+            echo "export BUILD_DIR=/tmp/build" >>$BASH_ENV
+            echo "export RESULT_DIR=/tmp/result" >>$BASH_ENV
+            echo "export CLOUDIFY_VERSION=5.1" >>$BASH_ENV
```
mabuaisha: @tehasdf cannot we generate the version in aumtoicat way to read it from setup.py instead of hard coded here, so every version we need to update that manually

.circleci/config.yml
```diff
@@ -84,7 +84,40 @@ jobs:
...
+            echo "export CLOUDIFY_PACKAGE_RELEASE=1" >>$BASH_ENV
+            mkdir /tmp/result
+      - run:
+          name: Set CLOUDIFY_VERSION
+          command: |
+            set -ex
+            apt-get update
+            apt-get install python python-setuptools -y
+            echo "export CLOUDIFY_VERSION=$(python ~/project/setup.py --version)" >>$BASH_ENV
```
mabuaisha: oh I noticed you are doing this here, should we then remove the above export ?

packaging/debian/build.sh
```diff
@@ -0,0 +1,41 @@
...
+virtualenv /opt/cfy
+/opt/cfy/bin/pip install -r "${PROJECT_DIR}/dev-requirements.txt"
+/opt/cfy/bin/pip install "${PROJECT_DIR}"
+cp /opt/cfy "${BUILD_DIR}/opt/cfy" -fr
+
+mkdir -p "${BUILD_DIR}/usr/bin"
+ln -s "/opt/cfy/bin/cfy" "${BUILD_DIR}/usr/bin/cfy"
+
+dpkg-deb --build "${BUILD_DIR}"
+mv ~/*.deb "${RESULT_DIR}"
```
mabuaisha: Do not you think we should also add/update a dockerfile so that we can use this on jenkins as well

.circleci/config.yml
tehasdf: oh whoops

packaging/debian/build.sh
tehasdf: dockerfiles aren't related, this uses docker anyway but via `docker exec`

PR #1074, opened 2020-03-24
CY-2436 package filename changes

tehasdf:
See commits for the .spec and the .deb change

### DIFF
--- a/packaging/debian/build.sh
+++ b/packaging/debian/build.sh
@@ -14,7 +14,7 @@ set -e
 }
 set -xu
 
-BUILD_DIR=~/cloudify-cli_${CLOUDIFY_VERSION}~${CLOUDIFY_PACKAGE_RELEASE}_amd64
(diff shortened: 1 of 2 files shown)


packaging/debian/build.sh
```diff
@@ -14,7 +14,7 @@ set -e
 }
 set -xu
 
-BUILD_DIR=~/cloudify-cli_${CLOUDIFY_VERSION}~${CLOUDIFY_PACKAGE_RELEASE}_amd64
+BUILD_DIR=~/cloudify-cli_${CLOUDIFY_VERSION}-${CLOUDIFY_PACKAGE_RELEASE}_amd64
```
alexmolev: I think the right thing will be to remove `-`
```suggestion
BUILD_DIR=~/cloudify-cli_${CLOUDIFY_VERSION}${CLOUDIFY_PACKAGE_RELEASE}_amd64
```

PR #1095, opened 2020-06-30
CY-2209 Add CLI command executions delete

glukhman:


cloudify_cli/commands/executions.py
```diff
@@ -484,3 +484,41 @@ def summary(target_field, sub_field, logger, client, tenant_name,
...
+        logger.critical('Must specify one of the options --keep-last | '
+                        '--keep-days | --keep-since')
+        return
+    deleted_executions = client.executions.delete(
+        keep_last=keep_last,
+        keep_days=keep_days,
+        keep_since=keep_since,
+        all_tenants=all_tenants,
+    )
+    logger.info('Successfully deleted {0} executions'.format(
```
tehasdf, line 523: not really important, but a pet peeve:
`logger.info('xxx {0}'.format(x))` <-- does the string interpolation no matter if the log is going to be shown or not, so it always does an allocation and costs performance
`logger.info('xxx %s', x)` <--- the logging framework only does the interpolation when the log is going to be actually shown

cloudify_cli/commands/executions.py
```diff
@@ -484,3 +484,41 @@ def summary(target_field, sub_field, logger, client, tenant_name,
         items,
         'Execution summary by {field}'.format(field=target_field),
     )
+
+
+@executions.command(name='delete',
+                    short_help='Delete finished executions from listing')
```
tehasdf, line 490: `from listing`?

cloudify_cli/commands/executions.py
```diff
@@ -484,3 +484,41 @@ def summary(target_field, sub_field, logger, client, tenant_name,
...
+@click.option('-n', '--keep-last', type=click.IntRange(min=1))
+@click.option('-d', '--keep-days', type=click.IntRange(min=1))
+@click.option('-s', '--keep-since', type=click.DateTime(['%Y-%m-%d']))
```
tehasdf, line 493: plz use the same format as events delete (to_datetime etc)

PR #1098, opened 2020-07-12
CY-3107 Implement replace-certificates start

YoniItzhak:


cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,273 @@
...
+                      config_dict['manager']['cluster_members'],
+                      force, logger)
+    _validate_ca_cert(errors_list,  config_dict['postgresql_server'],
+                      'postgresql_server', 'new_ca_cert_path',
+                      'new_postgresql_client_cert_path',
+                      config_dict['manager']['cluster_members'],
+                      force, logger)
+
+
+def _validate_ca_cert(errors_list, instance, instance_name, new_ca_cert_name,
```
tehasdf, line 211: it's always a good idea, but especially this function could use a docstring saying what is it actually validating
it seems that this checks that if you change the CA, you must change the corresponding cert for all nodes?

cloudify_cli/replace_certificates_config.py
```diff
@@ -0,0 +1,377 @@
...
+        client.connect(hostname=hostname, username=username,
+                       key_filename=key_file)
+        return client
+
+    def _create_ssh_client(self):
+        return retry_with_sleep(self._create_ssh_client_func,
+                                self.host_ip, self.username,
+                                self.key_file_path)
+
+    def _blocking_exec_command(self, command):
```
tehasdf, line 87: are you sure we want to use this, and not just reuse existing fabric stuff, this seems pretty low-level for what we want to do

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,273 @@
...
+                                  instances_ips['db_nodes_ips'])
+    _add_nodes_to_config_instance(config, 'rabbitmq',
+                                  instances_ips['broker_nodes_ips'])
+    return config
+
+
+def _add_manager_nodes_to_config_instance(config, manager_ips):
+    for host_ip in manager_ips:
+        instance = {'host_ip': str(host_ip),
+                    'new_internal_cert_path': '',
```
tehasdf, line 115: ALL of them (here and certs_mapping etc) start with `new` and end with `path`, so i'd just drop those pre/postfixes and named it just `internal_cert` instead. Makes it a bit easier on the eyes when reading.

cloudify_cli/replace_certificates_config.py
```diff
@@ -0,0 +1,377 @@
...
+                self.put_file(new_cert_path, remote_cert_path)
+
+        self._pass_config_file()
+
+    def _pass_config_file(self):
+        local_path = LOCAL_CONFIG_FILE_PATH.format(self.host_ip + '_')
+        with open(local_path, 'w') as config_file:
+            yaml.dump(self.certs_dict, config_file)
+
+        self.put_file(local_path, REMOTE_CONFIG_FILE_PATH)
```
tehasdf, line 194: fabric would let you put a stringio! :P

cloudify_cli/replace_certificates_config.py
```diff
@@ -0,0 +1,377 @@
...
+        self._pass_certificates()
+        self.run_command('cfy_manager replace-certificates -i {0}'
+                         ' --only-validate'.format(REMOTE_CONFIG_FILE_PATH))
+
+    def needs_to_replace_certificates(self):
+        return any(self.certs_mapping.values())
+
+    def _pass_certificates(self):
+        self._prepare_new_certs_dir()
+        for cert_name, new_cert_path in self.certs_mapping.items():
```
tehasdf, line 181: (not commenting about this line specifically but this line is the main path i guess)

I very much dislike the class-based stuff here, because all the classes take paths, and for the most part pass them through unchanged. I think the classes could be removed because they don't add much value.
The point is, to follow what happens to eg. `new_internal_cert_path`:
 - it is created in the config dict & file as `new_internal_cert_path`
 - it is loaded into `ReplaceCertificatesConfig`
 - `ReplaceCertificatesConfig` passes it - by name - into `ManagerNode`
 - `ManagerNode` receives it by name and sets it (well i gues the base class does) as `self.new_cert_path`
 - `_create_paths_mapping` runs and renames it back
 - the path is retrieved by name from `CERTS_MAPPING`
 - it is uploaded
 - it is put into the remote config file

see how many times i had to read `new_internal_cert_path`?! And there's like 10 of these files, which makes it hard to follow, because man my eyes start floating halfway deep into the stack. If somewhere you made a mistake and did `self.new_internal_cert = new_internal_key` then no way I'm gonna catch it :)
It's just reading the same name over and over but once in a while you throw a curveball and rename `new_internal_cert_path` to `new_cert_path` :D
(it doesn't help that all the names start with `new` and end with `path`, so they're more similar)

The main problem is that i must now hold several structures in my head: the config structure, the arguments to the Node classes, the dict output of the Node classes, the certs mapping, and the structure of the resulting remote config file.

I suggest reducing that count to just one item: the node config from the local config (the 
 item from `cluster_members`).
In there you have dicts with the following structure (example):
```python
{
    'new_internal_key_path': '',
    'new_external_cert_path': '',
    'new_external_key_path': '',
    'new_postgresql_client_cert_path': '',
    'new_postgresql_client_key_path': ''
}
```
So let's call a function `def upload_new_certs(host_ip, that_object):` which would: (pseudocode)
 - iterate over that dict, eg key=`new_internal_key_path`, value=`file.pem`
 - `conn = make_connection(host_ip)`
 - `local_path = config[key]` -
 - `remote_path = '/tmp/certs/{0}.pem'.format(key)` <- yes just that, no mapping
 - alternatively: `remote_path = '/tmp/certs/' + basename(local_path))` (same filename as local, doesnt really matter)
 - `conn.put_file(local_path, remote_path)`
 - put into the remote config file, `config[key] = remote_path` (no mapping)

it can still be written with classes if you want, or you can have several functions if you need specialcasing per node type, but the point is using the input config dict _directly_, without the N transformations to N object structures

cloudify_cli/replace_certificates_config.py
```diff
@@ -0,0 +1,377 @@
...
+class ReplaceCertificatesConfig(object):
+    def __init__(self, config_dict, is_all_in_one, logger):
+        if SSHClient is None:
+            raise CloudifyCliError("SSH not available - fabric not installed")
+
+        self.logger = logger
+        self.is_all_in_one = is_all_in_one
+        self.config_dict = config_dict
+        self.username = env.profile.ssh_user
+        self.key_file_path = env.profile.ssh_key
```
tehasdf, line 287: those can be hopefully made to use the cluster profile (in case of different names/keys per node)

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,273 @@
...
+                                         logger,
+                                         client):
+    # TODO: Take care of the AIO case
+    output_path = output_path if output_path else CERTS_CONFIG_PATH
+    config = _get_cluster_configuration_dict(client)
+    with open(output_path, 'w') as output_file:
+        yaml.dump(config, output_file, default_flow_style=False)
+
+    logger.info('The certificates replacement configuration file was '
+                'saved to {0}'.format(output_path))
```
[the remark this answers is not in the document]
YoniItzhak: Great point! I forgot about it and modified all places in the code that use it.

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,273 @@
...
+    Handle the certificates replacement procedure
+    """
+    if not env.is_initialized():
+        env.raise_uninitialized()
+
+
+@replace_certificates.command(name='generate-config',
+                              short_help='Generate the configuration file '
+                                         'needed for certificates replacement')
+@cfy.options.output_path
```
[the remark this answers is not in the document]
YoniItzhak: Oh I haven't thought of this option, I think I'll still leave it as a file since it might be easier for the user (I think)

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,273 @@
...
+                               force,
+                               logger):
+    # TODO: implement for the AIO case
+    # TODO: Validate the user role = admin
+    _validate_username_and_private_key()
+    config_dict = get_dict_from_yaml(_get_input_path(input_path))
+    logger.info('Validating replace-certificates config file...')
+    errors_list = validate_config_dict(config_dict, force, logger)
+    if errors_list:
+        raise_errors_list(errors_list, logger)
```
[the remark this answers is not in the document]
YoniItzhak: Great point! I'll fix it in all places.

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,273 @@
...
+
+    main_config = ReplaceCertificatesConfig(config_dict, False, logger)
+    are_certs_valid = main_config.validate_certificates()
+    if are_certs_valid:
+        logger.info('Replacing certificates')
+        replace_certs_succeeded = main_config.replace_certificates()
+        new_cli_cert = main_config.new_cli_ca_cert()
+        if replace_certs_succeeded and new_cli_cert:
+            os.system('cfy profiles set -c {0}'.format(new_cli_cert))
```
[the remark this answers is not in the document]
YoniItzhak: Right! I just updated the env.profile.rest_certificate

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,273 @@
...
+    logger.info('Validating replace-certificates config file...')
+    errors_list = validate_config_dict(config_dict, force, logger)
+    if errors_list:
+        raise_errors_list(errors_list, logger)
+
+    main_config = ReplaceCertificatesConfig(config_dict, False, logger)
+    are_certs_valid = main_config.validate_certificates()
+    if are_certs_valid:
+        logger.info('Replacing certificates')
+        replace_certs_succeeded = main_config.replace_certificates()
```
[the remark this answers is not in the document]
YoniItzhak: Right, changed :)

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,273 @@
...
+    if not os.path.exists(input_path):
+        raise CloudifyCliError('Please create the replace-certificates '
+                               'configuration file first using the command'
+                               ' `cfy replace-certificates generate-file`')
+    return input_path
+
+
+def _get_cluster_configuration_dict(client):
+    instances_ips = _get_instances_ips(client)
+    config = {}
```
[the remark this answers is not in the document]
YoniItzhak: Makes sense, thanks, I changed it so all the config will be created in the function.

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,273 @@
...
+                             'new_external_key_path')
+        _validate_node_certs(errors_list, node,
+                             'new_postgresql_client_cert_path',
+                             'new_postgresql_client_key_path')
+
+
+def _validate_node_certs(errors_list, certs_dict, new_cert_name, new_key_name):
+    new_cert_path = certs_dict.get(new_cert_name)
+    new_key_path = certs_dict.get(new_key_name)
+    if bool(new_key_path) != bool(new_cert_path):
```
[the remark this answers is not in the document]
YoniItzhak: Haha yeah, I think I've used it before, a great way to do xor :)

cloudify_cli/commands/replace_certificates.py
YoniItzhak: Correct, added a docstring

cloudify_cli/replace_certificates_config.py
YoniItzhak: Changed back to fabric. The purpose of this function is to print the output as it is printed at the remote host, but with fabric it's automatic.

cloudify_cli/commands/replace_certificates.py
YoniItzhak: Makes a lot of sense, I dropped the `new` and kept the `path` to be consistent with the config.yaml file

cloudify_cli/replace_certificates_config.py
YoniItzhak: Nice! I tested it now and it seems that StringIO doesn't except dictionaries as inputs?

cloudify_cli/replace_certificates_config.py
YoniItzhak: Correct :)

cloudify_cli/replace_certificates_config.py
YoniItzhak: This is so true! Thanks, I also thought this way was too complicated but couldn't find a way to simplify it. Changed it now.

cloudify_cli/replace_certificates_config.py
tehasdf: @YoniItzhak stringio is a file-like object, just in memory instead of on disk (so you dont need to remove it etc). You still write to it, same as with a file, so `yaml.dump(xxx, some_stringio)` should work

YoniItzhak: @tehasdf Thanks for the explanation! I'm not using the file anymore so it's fine :)

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,251 @@
...
+                                         'needed for certificates replacement')
+@cfy.options.output_path
+@cfy.assert_manager_active()
+@cfy.pass_client()
+@cfy.pass_logger
+def get_replace_certificates_config_file(output_path,
+                                         logger,
+                                         client):
+    # TODO: Take care of the AIO case
+    output_path = output_path if output_path else CERTS_CONFIG_PATH
```
geokala, line 52: Is there a reason this isn't a default argument?

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,251 @@
...
+                              short_help='Generate the configuration file '
+                                         'needed for certificates replacement')
+@cfy.options.output_path
+@cfy.assert_manager_active()
+@cfy.pass_client()
+@cfy.pass_logger
+def get_replace_certificates_config_file(output_path,
+                                         logger,
+                                         client):
+    # TODO: Take care of the AIO case
```
geokala, line 51: Can we either do the various TODOs, or mark in comments that they're not done and put a jira ID as a reference?

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,251 @@
...
+    main_config = ReplaceCertificatesConfig(config_dict, False, logger)
+    main_config.validate_certificates()
+    main_config.replace_certificates()
+    new_cli_cert = main_config.new_cli_ca_cert()
+    if new_cli_cert:
+        env.profile.rest_certificate = new_cli_cert
+
+
+def _get_input_path(input_path):
+    input_path = input_path if input_path else CERTS_CONFIG_PATH
```
geokala, line 88: This looks like a default arg too?

cloudify_cli/commands/replace_certificates.py
alexmolev, line 51: Please create a ticket in Jira for it and remove TODO.

cloudify_cli/commands/replace_certificates.py
```diff
@@ -0,0 +1,251 @@
...
+                                         'the configuration file')
+@cfy.options.input_path(help='The certificates replacement configuration file')
+@cfy.options.force('Use the force flag in case you want to change only a '
+                   'CA and not the certificates signed by it')
+@cfy.assert_manager_active()
+@cfy.pass_logger
+def start_replace_certificates(input_path,
+                               force,
+                               logger):
+    # TODO: implement for the AIO case
```
alexmolev, line 72: Same here

cloudify_cli/commands/replace_certificates.py
YoniItzhak: Thanks, done.

YoniItzhak: Sure, done.

cloudify_cli/commands/replace_certificates.py
YoniItzhak: Well, only if the user wants to save it to somewhere else

cloudify_cli/commands/replace_certificates.py
YoniItzhak: Same as the above.

cloudify_cli/commands/replace_certificates.py
geokala: Right, but rather than putting:
def func(something):
    something = something is something else THING

You could just put:
def func(something=THING):

YoniItzhak: Right!

PR #1102, opened 2020-07-19
CY-3149 Update locale for python 3

mabuaisha:


cloudify_cli/cli/cfy.py
```diff
@@ -455,7 +459,78 @@ def pass_context(func):
...
-class AliasedGroup(click.Group):
+class CommandMixin(object):
+    """
+    This class mixin helps to set the right locale for system required
+    by python 3 for click library where "LC_ALL" & "LANG" are not set and
+    in order to avoid the RuntimeError raised by click library which
+    prevents invoking cfy commands
+    """
+
+    def __call__(self, *args, **kwargs):
```
geokala, line 470: Can we remove this method? It's effectively a 5 line NOP.

mabuaisha: Dropped

PR #1103, opened 2020-07-20
CY-3111 Replace certificates on agents

YoniItzhak:


cloudify_cli/replace_certificates_config.py
```diff
@@ -136,10 +137,25 @@ def _validate_certificates(self):
         self._close_clients_connection()
 
     def replace_certificates(self):
+        self._pass_new_ca_certs_to_agents(bundle=True)
```
geokala, line 140: I'm not sure I understand why you pass them as a bundle then as not a bundle- can you comment this?

YoniItzhak: Sure, so I first pass them as a bundle to make sure the old+new certs are there. Then, at the end of the process, I'm only leaving the new ones in order to prevent security issues.

PR #1122, opened 2020-08-10
CY-3302 Implement `cfy plugins install`

tehasdf:
Implement the command, and some tests.

Also, make some other code in that module better.

### DIFF
--- a/cloudify_cli/commands/plugins.py
+++ b/cloudify_cli/commands/plugins.py
@@ -15,12 +15,18 @@
 ############
 
 import os
+import time
 
+import click
 import wagon
+
+from cloudify.models_states import PluginInstallationState
 
 from cloudify_cli import execution_events_fetcher
 from cloudify_cli.logger import get_events_logger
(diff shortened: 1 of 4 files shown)


cloudify_cli/commands/plugins.py
```diff
@@ -215,6 +221,93 @@ def get(plugin_id, logger, client, tenant_name, get_data):
...
+                if pstate.get('manager') in wait_managers:
+                    errors += 1
+                    wait_managers.remove(pstate['manager'])
+                    logger.info('Error installing on manager %s: %s',
+                                pstate['manager'], pstate['error'])
+                if pstate.get('agent') in wait_agents:
+                    errors += 1
+                    wait_agents.remove(pstate['agent'])
+                    logger.info('Error installing on agent %s: %s',
+                                pstate['agent'], pstate['error'])
```
tehasdf, line 258: I'm not turbo happy with this code but I haven't been able to structure it better :(

PR #1124, opened 2020-08-11
CY-3316 Fix the prometheus case

YoniItzhak:


YoniItzhak:
This PR fixes a few things in the `cfy certificates replace` command:
1. Replacing the CLI and client certificates at the end of the process.
2. Adding a retry to the status validation at the end of the process, since the cluster services might take time to restart.

PR #1128, opened 2020-08-17
CY-3328 Fix replace-certificates error handling

YoniItzhak:


YoniItzhak:
I implemented a few changes: 
1. Added a verbose flag to print the logs from the remote hosts. `-v` prints the only the `INFO` and `ERROR` logs. `-vv` and `-vvv` print all logs. 
2. Dropped the `self.connection` attr and replaced it with a context-manager. This allowed me to also drop `_close_clients_connection`.

cloudify_cli/replace_certificates_config.py
```diff
@@ -39,46 +43,65 @@ def __init__(self,
...
+            host=self.host_ip, user=self.username, port=22,
+            connect_kwargs={'key_filename': self.key_file_path})
 
-        except AuthenticationException as e:
-            self.errors_list.append(
-                "SSH: could not connect to {host} "
-                "(username: {user}, key: {key}): {exc}".format(
-                    host=self.host_ip, user=self.username,
-                    key=self.key_file_path, exc=e))
+    def _raise_authentication_err(self, exc):
```
tehasdf, line 55: this doesn't raise (anymore) so probably change the name to not include `raise`

PR #1129, opened 2020-08-18
Adding fabric to the CLI requirements

YoniItzhak:


YoniItzhak:
* Adding fabric to the requirements as it should not break anything. 
* Adding `expanduser` in case the path includes `~`.

PR #1131, opened 2020-08-23
CY-3263 Update cfy cluster commands

YoniItzhak:


tehasdf (approved):

YoniItzhak:
In v5.1 we dropped the `node_id` attribute, therefore it shouldn't (can't) be used in the CLI code.

PR #1132, opened 2020-08-30
Fixing fabric error handling in replace certificates

YoniItzhak:


YoniItzhak:
This PR fixes two things:
1. Encoding the error message so we will be able to print special characters.
2. Trying to open the connection before it is returned to the `run_command` and `put_file` commands. This will verify the connection before the function tries to run.

tehasdf (approved):

PR #1134, opened 2020-09-06
CY-3436 Add tenant name to cfy agents commands

YoniItzhak:


YoniItzhak:
This PR serves a few objectives:
1. Adding `tenant-name` to `cfy agents list`, enabling users to list agents of a specific deployment. 
2. Logging `Explicitly using tenant <tenant-name>` in case of using `cfy agents <validate, install, list> --tenant-name <tenant-name>` command. This will make the code more explicit and prevent confusion of whether the tenant-name is in use or not (like happened to me).

geokala:
Do you mean "enabling users to list agents of a specific tenant" or do you mean that you added --deployment to cfy agents list?

geokala (approved):

geokala (changes requested):
As pointed out in chat by @tehasdf , tenant name should probably be used (I assumed the decorator was doing something magic).

geokala (approved):
OK, apparently the pass_client decorator handles this in some way, so approving again.

PR #1136, opened 2020-09-23
CY-3552 Use b64encode for creating credentials.

tehasdf:
As opposed to urlsafe_b64encode.

No reason for this to be urlsafe. At all.

Also, urlsafe breaks, because flask uses the non-urlsafe version
when decoding: https://github.com/pallets/werkzeug/blob/bbc5cbb8390d8d9de4fff42456cab3d91b85b584/src/werkzeug/http.py#L622
(this is for making `request.authorization`, which we then use).
Also, the basic auth rfc (RFC 7617) talks about using the plain version
of base64, not an urlsafe variant.

It needs to be encoded the same way it's going to be decoded.

### DIFF
--- a/cloudify_cli/env.py
+++ b/cloudify_cli/env.py
@@ -21,7 +21,7 @@ import shutil
 import getpass
 import tempfile
 import itertools
-from base64 import urlsafe_b64encode
+from base64 import b64encode
 from contextlib import contextmanager
 
 import yaml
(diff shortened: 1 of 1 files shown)


tehasdf:
The rest-client counterpart to this is cloudify-cosmo/cloudify-common#557

geokala (approved):

PR #1142, opened 2020-10-07
CY-3638 agent install: fetch all node-instances

tehasdf:
Instead of limiting the node-instance list by deployment id,
just fetch all of them.

We can't put the list of deployments in the filters, because
those filters go into the rest call URL, and with thousands of
deployments, we easily hit the URL size limit.

Instead, let's just fetch _all_ the node instances. That approach
is obviously correct, even if wasteful.

It is "wasteful" because we will be fetching node instances even
for deployments that we have already filtered out by using
the --node-id filter.
This means that this approach is going to be doing unnecessary
requests, and will be slower, for `cfy agents install --node-id x`.

However, using --node-id alone, without --deployment-id is
a very rare, and not a very useful, usecase, that's why I'm not
worried about making it slower.
(and anyway, it is going to be slower by the time it takes to
fetch node-instances, which pales in comparison to the time it takes
to actually run the agents install!)

Also, we used to only do a single request for the node-instances,
without following pagination. That's a separate bug in itself,
because it meant if there was more than one page worth of instances,
we would skip some.
Now we do follow pagination.

### DIFF
--- a/cloudify_cli/commands/agents.py
+++ b/cloudify_cli/commands/agents.py
@@ -226,28 +226,35 @@ def get_filters_map(
             for tenant_name in list(tenants_to_deployments):
                 tenant_client = env.get_rest_client(tenant_name=tenant_name)
                 deps_to_execute = tenants_to_deployments[tenant_name]
+                offset = 0
+                while True:
+                    node_instances = tenant_client.node_instances.list(
+                        _include=['id', 'host_id', 'deployment_id', 'state'],
+                        _offset=offset,
+                    )
+                    # Find all unstarted Compute instances.
+                    unstarted_computes = [
+                        ni for ni in node_instances
+                        if ni.id == ni.host_id and
+                        ni.state != _NODE_INSTANCE_STATE_STARTED
+                    ]
 
-                node_instances = tenant_client.node_instances.list(
-                    deployment_id=list(deps_to_execute),
-                    _include=['id', 'host_id', 'deployment_id', 'state']
-                )
+                    for unstarted_ni in unstarted_computes:
+                        logger.info("Node instance '%s' is not in '%s' state; "
+                                    "deployment '%s' will be skipped",
+                                    unstarted_ni.id,
+                                    _NODE_INSTANCE_STATE_STARTED,
+                                    unstarted_ni.deployment_id)
+                        deps_to_execute.pop(unstarted_ni.deployment_id, None)
 
-                # Find all unstarted Compute instances.
-                unstarted_computes = [
-                    ni for ni in node_instances
-                    if ni.id == ni.host_id and
-                    ni.state != _NODE_INSTANCE_STATE_STARTED
-                ]
+                    if not deps_to_execute:
+                        del tenants_to_deployments[tenant_name]
 
-                for unstarted_ni in unstarted_computes:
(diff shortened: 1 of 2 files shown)


tehasdf:
Also, added offset handling to the unittest mocks, and added a unittest. See each commit for details.

tehasdf:
The PR message is the same as the first commit message, so you don't need to read both of these :)

cloudify_cli/commands/agents.py
```diff
@@ -226,28 +226,35 @@ def get_filters_map(
...
-                    logger.info("Node instance '%s' is not in '%s' state; "
-                                "deployment '%s' will be skipped",
-                                unstarted_ni.id, _NODE_INSTANCE_STATE_STARTED,
-                                unstarted_ni.deployment_id)
-                    deps_to_execute.pop(unstarted_ni.deployment_id, None)
-
-                if not deps_to_execute:
-                    del tenants_to_deployments[tenant_name]
+                offset = 0
+                while True:
```
tehasdf, line 230: i've put the whole thing under this `while` (which is stopped by the `break` later), because this loop is for following pagination

cloudify_cli/commands/agents.py
```diff
@@ -226,28 +226,35 @@ def get_filters_map(
...
+                                    _NODE_INSTANCE_STATE_STARTED,
+                                    unstarted_ni.deployment_id)
+                        deps_to_execute.pop(unstarted_ni.deployment_id, None)
+
+                    if not deps_to_execute:
+                        del tenants_to_deployments[tenant_name]
+
+                    size = node_instances.metadata.pagination.size
+                    total = node_instances.metadata.pagination.total
+                    if len(node_instances) < size or size == total:
```
tehasdf, line 255: the last request is going to have less than a full page worth of instances (`len() < size`), that's when we know to stop paginating, because there's no more instances

alternatively, if the whole thing fits on a single page (`size == total`) then we know to not even bother fetching a second page

cloudify_cli/commands/agents.py
```diff
@@ -226,28 +226,35 @@ def get_filters_map(
             for tenant_name in list(tenants_to_deployments):
                 tenant_client = env.get_rest_client(tenant_name=tenant_name)
                 deps_to_execute = tenants_to_deployments[tenant_name]
-
-                node_instances = tenant_client.node_instances.list(
-                    deployment_id=list(deps_to_execute),
```
tehasdf, line 231: this is the important part

cloudify_cli/commands/agents.py
```diff
@@ -226,28 +226,35 @@ def get_filters_map(
...
+                    for unstarted_ni in unstarted_computes:
+                        logger.info("Node instance '%s' is not in '%s' state; "
+                                    "deployment '%s' will be skipped",
+                                    unstarted_ni.id,
+                                    _NODE_INSTANCE_STATE_STARTED,
+                                    unstarted_ni.deployment_id)
+                        deps_to_execute.pop(unstarted_ni.deployment_id, None)
+
+                    if not deps_to_execute:
+                        del tenants_to_deployments[tenant_name]
```
tehasdf, line 251: this didn't change, it was just indented into under the loop

mateumann (approved):

alexmolev (approved):

PR #1145, opened 2020-10-11
Cy 3105 migrate jenkinsfile

yuvlar:
Migrating circleci config to jenkins jenkinsfile


jenkins/Jenkinsfile
```diff
@@ -0,0 +1,178 @@
...
+                  flake8 cloudify_cli
+                '''
+              }
+            }
+          }
+        }
+        stage('build-rpm') {
+          steps {
+            container('rpmbuild') {
+              sh """
```
warolv, line 82: why """, use ''' here - I see you don't need interpolation

warolv: also better to combine with 'Installing setup tools' shell or write echo before this one

jenkins/build-pod.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - name: debbuild
+    image: debian:9
+    command:
+    - cat
+    tty: true
+    securityContext:
+      runAsUser: 0
+      privileged: true
+  nodeSelector:
+  instance-type: spot-xlarge
```
warolv, line 47: how much time this flow runs on xlarge? and on large instance?

warolv (changes requested):

jenkins/build-pod.yaml
yuvlar: For the flow it takes about the same time, maybe a few seconds difference, the big difference is in creating the pod, with large instance it can take a while for it to load the pod, while with xlarge it's definitely quicker.
I changed it to xlarge mainly to have faster builds to test my changes faster.
I'll change and keep it on large, so it won't consume unneeded resources.

jenkins/Jenkinsfile
yuvlar: Thanks,
Changed

jenkins/build-pod.yaml
warolv: Sounds good

warolv (approved):
Good job!

PR #1152, opened 2020-10-19
cfy prof use --skip-cred-val: avoid http requests

tehasdf:
The point of `--skip-credentials-validation` is to not run requests
against the manager, so we should avoid doing that.
Updating the cluster profile does need to run requests, to download
the list of managers/etc


geokala:

PR #1161, opened 2020-11-18
RD-347 Fix cfy logs command

glukhman:


glukhman:
1. change `archive_path` composition to support usage with Windows CLI
2. Fix for loop in purge: `sudo for...` is invalid bash syntax

YoniItzhak (approved):

PR #1162, opened 2020-11-24
RD-347 Add JSON output to logs download

glukhman:


glukhman:
This is for improved access to the paths of the downloaded archives by automated tools (e.g. system tests)

cloudify_cli/commands/logs.py
```diff
@@ -72,6 +74,7 @@ def _download_archive(conn, host_type, output_path, logger):
     conn.get(archive_path_on_host, output_path)
     logger.info('Removing archive from host...')
     conn.sudo('rm {0}'.format(archive_path_on_host))
+    output_json['archive paths'][conn.host] = output_path
```
YoniItzhak, line 77: I don't think this would work in the 3-nodes case. You would end up having 3 archive paths instead of 9.

cloudify_cli/commands/logs.py
```diff
@@ -83,6 +86,8 @@ def _download_archive(conn, host_type, output_path, logger):
 def download(output_path, all_nodes, logger):
     """Download an archive containing all of the manager's service logs
     """
+    output_json = {'archive paths': {}}
```
YoniItzhak, line 89: A suggestion. Instead of passing this dictionary to the `_download_archive` function, make the `_download_archive` function return the `output_path` and use it in this function. 
I, personally, don't like modifying the same dictionary in different functions.

cloudify_cli/commands/logs.py
```diff
@@ -110,7 +116,11 @@ def download(output_path, all_nodes, logger):
                 conn,
                 CloudifyNodeType.MANAGER,
                 output_path,
-                logger)
+                logger,
+                output_json)
+
+    if get_global_json_output():
+        output(json.dumps(output_json))
```
YoniItzhak, line 123: So your system test would parse the stdout of the function?

YoniItzhak: You might want to use the `cls` parameter like here https://github.com/cloudify-cosmo/cloudify-cli/blob/master/cloudify_cli/commands/cluster.py#L173

YoniItzhak (approved):

PR #1186, opened 2021-01-17
RD-1054 Add exec schedules CLI

glukhman:


cloudify_cli/commands/executions.py
```diff
@@ -526,3 +532,318 @@ def delete(logger, client, tenant_name, to_datetime, before, keep_last,
...
+                    tenant_name,
+                    client,
+                    logger):
+    """Schedule an execution of a workflow on a given deployment
+    """
+    utils.explicit_tenant_name_message(tenant_name, logger)
+    logger.info('Scheduling the execution of workflow `{0}` on deployment '
+                '`{1}`. Schuedule name: {2}'.format(workflow_id,
+                                                    deployment_id,
+                                                    name))
```
tehasdf, line 595:
```suggestion
    logger.info('Scheduling the execution of workflow `%s` on deployment '
                '`%s`. Schuedule name: %s',
                workflow_id, deployment_id, name))
```
the usual str.format + logging comment
(multiple occurences in this pr)

cloudify_cli/commands/executions.py
```diff
@@ -526,3 +532,318 @@ def delete(logger, client, tenant_name, to_datetime, before, keep_last,
...
+                     to_datetime,
+                     freq,
+                     count,
+                     weekdays,
+                     rrule,
+                     tenant_name,
+                     pagination_size,
+                     client,
+                     logger):
+    """Preview the next running dates for a schedule. Works either with an
```
tehasdf, line 700: it would be good if we were able to push the computation to the server-side, beause now we have two places that do this, and good luck making sure they're the same!

also this requires us to have dateutil and it would be good if we didnt have to add new dependencies to the cli

tehasdf (approved):

PR #1199, opened 2021-02-02
RD-1284 Stop waiting for deployment environment creation when trigger…

mabuaisha:
This PR created in order to stop waiting the execution for deployment as we are no longer have a workflow that create deployment directory on the file server whenever we create a deployment


geokala:

PR #1237, opened 2021-03-15
RD-1698 Deprecate the --schedule flag in cfy exec start

glukhman:


cloudify_cli/commands/executions.py
```diff
@@ -265,7 +271,9 @@ def manager_start(workflow_id,
                         ' start when possible.')
             return
         if execution.status == 'scheduled':
-            logger.info('Execution is scheduled for {0}.'.format(schedule))
+            logger.info('Execution is scheduled for {0}.\nYou can see the '
+                        'execution schedule using `cfy deployments schedule '
+                        'list`.'.format(schedule))
```
tehasdf, line 276: aww this the only place you used logger+str.format :P (yes I know it was like this before)

tehasdf (approved):

PR #1262, opened 2021-03-31
RD-1902 Bump cryptography from 3.3.1 to 3.4.7

mateumann:


mateumann:
must stick with 3.3.2 because of Python 2.7 in other repositories

PR #1264, opened 2021-03-31
RD-1693 Allow special chars in labels

YoniItzhak:
If we want to allow special characters in labels, we need to modify the way they are currently provided in the CLI. This PR includes a few changes:

- Modifying the way we parse a provided labels list. This PR allows passing special characters in labels' values, where `,` and `:` must be escaped with `\`. 
- Modifying the way a user supplies filter rules. Following this PR, instead of supplying a string of filter rules separated by an ` and `, the user supplies each filter rule separately. This is done because I couldn't figure out a nice way to separate the big string of filter rules. 
- Moreover, the filter rules will be parsed using regular expressions to make sure we capture the right parts of each filter rule.

Associated PR: https://github.com/cloudify-cosmo/cloudify-manager/pull/2847

### DIFF
--- a/cloudify_cli/cli/cfy.py
+++ b/cloudify_cli/cli/cfy.py
@@ -10,6 +10,7 @@ import subprocess
 import locale
 import codecs
 from functools import wraps
+import unicodedata
 
 import click
 
@@ -256,13 +257,16 @@ def parse_and_validate_label_to_delete(ctx, param, value):
     return label_key
 
 
-def get_formatted_labels_list(raw_labels_list, allow_only_key=False):
+def get_formatted_labels_list(raw_labels_string, allow_only_key=False):
     labels_list = []
-    if '\x00' in raw_labels_list:
-        raise CloudifyValidationError('Error: labels cannot contain null')
+    if any(unicodedata.category(char)[0] == 'C' or char == '"'
+           for char in raw_labels_string):
+        raise CloudifyValidationError(
+            'Error: labels cannot contain control characters or `"`')
+
     format_err_msg = 'Labels should be of the form <key>:<value>,<key>:<value>'
-    raw_labels_list = raw_labels_list.replace('\\,', '\x00').split(',')
-    for label in raw_labels_list:
+    raw_labels_string = raw_labels_string.replace('\\,', '\x00').split(',')
+    for label in raw_labels_string:
         label = label.replace('\x00', ',')
         label = label.replace('\\:', '\x00')
         colons_count = label.count(':')
@@ -276,10 +280,6 @@ def get_formatted_labels_list(raw_labels_list, allow_only_key=False):
             if not label_key or not label_value:
                 raise LabelsValidationError(label, format_err_msg)
             label_value = label_value.replace('\x00', ':')
-            if any(char in label_value for char in ['"', '\n', '\t']):
-                raise LabelsValidationError(
-                    label_value,
-                    "The label's value contains one of `\"`, `\\n`, `\\t`")
 
         else:
             if allow_only_key:
--- a/cloudify_cli/cli/cfy.py
+++ b/cloudify_cli/cli/cfy.py
@@ -26,7 +26,9 @@ from ..inputs import inputs_to_dict
 from ..utils import generate_random_string
 from ..constants import DEFAULT_BLUEPRINT_PATH
 from ..exceptions import SuppressedCloudifyCliError
-from ..exceptions import CloudifyBootstrapError, CloudifyValidationError
+from ..exceptions import (LabelsValidationError,
+                          CloudifyBootstrapError,
+                          CloudifyValidationError,)
 from ..logger import (
     get_logger,
     set_global_verbosity_level,
@@ -233,19 +235,7 @@ def parse_and_validate_labels(ctx, param, value):
         raise CloudifyValidationError(
             'ERROR: The `{0}` argument is empty'.format(param.name))
 
-    labels_list = []
-    raw_labels_list = value.split(',')
-    for label in raw_labels_list:
-        if label.count(':') != 1:
-            raise CloudifyValidationError('ERROR: Labels should be of the '
-                                          'form <key>:<value>,<key>:<value>')
-
-        label_key, label_value = label.split(':')
-        validate_param_value('The key of one or more labels', label_key)
-        validate_param_value('The value of one or more labels', label_value)
-        labels_list.append({label_key: label_value})
-
-    return labels_list
+    return get_formatted_labels_list(value)
 
 
 def parse_and_validate_label_to_delete(ctx, param, value):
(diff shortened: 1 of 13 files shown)


cloudify_cli/cli/cfy.py
```diff
@@ -256,40 +246,77 @@ def parse_and_validate_label_to_delete(ctx, param, value):
...
-    return create_labels_filter_rules_list(value)
+        labels_list.append({label_key: label_value})
+
+    return labels_list
 
 
-def parse_attributes_filter_rules(ctx, param, value):
-    if value is None or ctx.resilient_parsing:
+def _validate_filter_rules_not_empty(ctx, param, value):
+    if value is None or value == () or ctx.resilient_parsing:
```
glukhman, line 305: why is the addition necessary?

cloudify_cli/cli/helptexts.py
```diff
@@ -499,33 +499,34 @@
...
-                     "<value>, <key> does-not-contain <value>, " \
+LABELS = "A labels list of the form <key>:<value>,<key>:<value>. " \
+         "Any comma and colon in <value> must be escaped with `\\`."
+
+LABELS_FILTER_RULES = "A labels' filter rule. Labels' filter rules must be " \
+                      "one of: <key>=<value>, <key>!=<value>, <key> is null," \
+                      " <key> is not null. <value> can be a single string " \
+                      "or a list of strings of the form " \
+                      "[<value1>,<value2>,...]. Any comma in <value> must " \
+                      "be escaped with `\\`. The labels' filter rules " \
```
glukhman, line 510: also colons, no?

cloudify_cli/filters_utils.py
```diff
@@ -36,7 +36,10 @@ def __init__(self, labels_filter_value):
             'The labels filter rule `{0}` is not in the right format. It must '
             'be one of: <key>=<value>, <key>!=<value>, <key> is null, '
             '<key> is not null. <value> can be a single string or a list of '
-            'strings of the form [<value1>,<value2>,...]'.format(
+            'strings of the form [<value1>,<value2>,...]. '
+            '<value> cannot contain `"`, `\\n` and `\\t`, plus, any comma in '
+            '<value> must be escaped with `\\`. <key> can contain only '
```
glukhman, line 41: as above, I'm confused why not colons as well

cloudify_cli/cli/helptexts.py
glukhman: Oh, so colons are escaped when providing labels, because of the k:v "assignment" form, but here we compare, not assign, so no need to escape colons. Umm... this is unfortunate. The user will have to be bothered to understand/remember this. I think better just always escape colons

cloudify_cli/filters_utils.py
glukhman: same as above

cloudify_cli/cli/cfy.py
```diff
@@ -256,40 +246,77 @@ def parse_and_validate_label_to_delete(ctx, param, value):
...
-        validate_param_value('The key of the provided label', label_key)
-        validate_param_value('The value of the provided label', label_value)
-        return {label_key: label_value}
 
+def get_formatted_labels_list(raw_labels_list, allow_only_key=False):
+    labels_list = []
+    if '\x00' in raw_labels_list:
+        raise CloudifyValidationError('Error: labels cannot contain null')
+    format_err_msg = 'Labels should be of the form <key>:<value>,<key>:<value>'
+    raw_labels_list = raw_labels_list.replace('\\,', '\x00').split(',')
```
glukhman, line 264: A general qualm: ASCII actually supports `\x1A` (Substitute character) for this. You can also use any of `\x1C` to `\x1F` and thus avoid the double substitution you had to make here

cloudify_cli/cli/cfy.py
```diff
@@ -256,40 +246,77 @@ def parse_and_validate_label_to_delete(ctx, param, value):
...
-        return label[0]
-    else:
-        label_key, label_value = label
-        validate_param_value('The key of the provided label', label_key)
-        validate_param_value('The value of the provided label', label_value)
-        return {label_key: label_value}
 
+def get_formatted_labels_list(raw_labels_list, allow_only_key=False):
+    labels_list = []
+    if '\x00' in raw_labels_list:
```
glukhman, line 261: I'd bar any ASCII lower than `\x20` really, they can easily mess things up (imagine how the label's display might look with them). But that of course depends of the customer's needs

cloudify_cli/labels_utils.py
```diff
@@ -46,10 +48,26 @@ def get_output_resource_labels(raw_resource_labels):
...
+                                  label_value in resource_label_values]
+        printable_labels.append({'key': resource_label_key,
+                                 'values': formatted_label_values})
+    return printable_labels
+
+
+def _format_label_value(label_value):
+    label_value = label_value.replace(',', '\\,').replace(':', '\\:').\
+        replace('$', '\\$')
+    if label_value_needs_quotes(label_value):
```
glukhman, line 63: Not sure why not just quote any label value, for uniformity's sake

cloudify_cli/cli/cfy.py
glukhman: Same for `\x7F` (DELETE)

cloudify_cli/filters_utils.py
```diff
@@ -49,128 +52,104 @@ def __init__(self, labels_filter_value):
...
-                return cls(key, [], OPERATOR_MAPPING['is null'])
-            elif match_not_null:
-                key = match_not_null.group(1).lower()
-                return cls(key, [], OPERATOR_MAPPING['is not null'])
-            else:
-                raise cls.format_exception(str_filter_rule)
-
-        else:
-            raise cls.format_exception(str_filter_rule)
+        match_equal = re.match(r'^([\w\-\.]+)(=)([^\n\t\"]+)$',
```
glukhman, line 94: You'll also need to escape = and !, come to think of it

cloudify_cli/cli/cfy.py
YoniItzhak: Oh, I wonder what does the word `substitute means there. In some places I read, the `\x1A` is used to specify EOF. And I see the point of using the Control characters for this purpose, it's a good idea. I'll also disallow them.

cloudify_cli/cli/cfy.py
YoniItzhak: Labels won't contain any control characters.

cloudify_cli/labels_utils.py
YoniItzhak: Agreed.

cloudify_cli/filters_utils.py
YoniItzhak: As explained on Slack, this is handled by the regexp, so need to escape `=` and `!`.

cloudify_cli/cli/helptexts.py
YoniItzhak: Agreed.

tehasdf (dismissed):

tehasdf (approved):

PR #1281, opened 2021-04-18
RD-2068 Fix delete sanity deployment with external db

YoniItzhak:
We should increase the timeout to allow the `delete_deployment_environment` workflow to finish when working with external DB.


tehasdf:
not for 5.1.2

PR #1307, opened 2021-05-16
RD-2283 Fix cfy logs download for 3 nodes cluster

mabuaisha:
This Fix added in order to allow generate logs archives based on the host ip + node type instead of depending on profile manager ip. Moreover, it adds support to validate if both ssh_user & ssh_user are set before downloading logs

### DIFF
--- a/cloudify_cli/commands/logs.py
+++ b/cloudify_cli/commands/logs.py
@@ -41,14 +41,14 @@ def get_host_date(conn):
     return conn.run('date +%Y%m%dT%H%M%S', hide=True).stdout.strip()
 
 
-def _archive_logs(conn, node_type, logger):
+def _archive_logs(conn, node_type, logger, node_ip):
     """Creates an archive of all logs found under /var/log/cloudify plus
     journalctl.
     """
     archive_filename = 'cloudify-{node_type}-logs_{date}_{ip}.tar.gz'.format(
         node_type=node_type,
         date=get_host_date(conn),
-        ip=env.profile.manager_ip
+        ip=node_ip
     )
     archive_path = '/tmp/{}'.format(archive_filename)
     journalctl_destination_path = '/var/log/cloudify/journalctl.log'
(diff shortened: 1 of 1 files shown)


cloudify_cli/commands/logs.py
```diff
@@ -93,6 +95,13 @@ def download(output_path, all_nodes, logger):
 
     if not output_path:
         output_path = os.getcwd()
+
+    _profile = env.profile
+    if not all([_profile.ssh_user, _profile.ssh_key]):
```
tehasdf, line 100: `all` for 2 items? lol maybe `and` :)

cloudify_cli/commands/logs.py
```diff
@@ -93,6 +95,13 @@ def download(output_path, all_nodes, logger):
...
+        raise CloudifyCliError(
+            '`ssh_user` & `ssh_key` must be set before '
+            'download logs'
```
tehasdf, line 103: grammar plz

cloudify_cli/commands/logs.py
```diff
@@ -93,6 +95,13 @@ def download(output_path, all_nodes, logger):
 
     if not output_path:
         output_path = os.getcwd()
+
+    _profile = env.profile
```
tehasdf, line 99: why set this?

cloudify_cli/commands/logs.py
```diff
@@ -41,22 +41,23 @@ def get_host_date(conn):
...
-def _archive_logs(conn, node_type, logger):
+def _archive_logs(conn, node_type, logger, node_ip=None):
     """Creates an archive of all logs found under /var/log/cloudify plus
     journalctl.
     """
     archive_filename = 'cloudify-{node_type}-logs_{date}_{ip}.tar.gz'.format(
         node_type=node_type,
         date=get_host_date(conn),
-        ip=env.profile.manager_ip
+        ip=node_ip or env.profile.manager_ip
```
tehasdf, line 51: probably better to just make this always take the ip, then it won't have the env.profile dependency

tehasdf (commented):
so why did this not work before?

cloudify_cli/commands/logs.py
mabuaisha: Fixed

cloudify_cli/commands/logs.py
mabuaisha: Updated

cloudify_cli/commands/logs.py
mabuaisha: Updated

cloudify_cli/commands/logs.py
mabuaisha: Done

tehasdf (approved):

PR #1319, opened 2021-06-23
RD-518 No need to wait for events after execution

tehasdf:
Events/logs are now sent in via REST, so it's synchronous wrt.
ordering vs the execution ending.
No need to wait the additional 3 seconds anymore!


mateumann (approved):

geokala (approved):
I'm very disappointed that these aren't my -lines.

PR #1322, opened 2021-07-05
RD-2705 Add ca_key to certs replacement config

glukhman:


cloudify_cli/commands/certificates.py
```diff
@@ -113,14 +113,19 @@ def _get_configuration_dict(client):
...
                 'new_prometheus_key': '',      # monitoring_service
                 'new_prometheus_ca_cert': '',  # was installed
+                'new_prometheus_ca_key': '',   # -------------
                 'new_ca_cert': '',
+                'new_ca_key': '',
                 'new_external_ca_cert': '',
-                'new_ldap_ca_cert': ''  # Relevant only if using LDAP
+                'new_external_ca_key': '',
+                'new_ldap_ca_cert': '',  # Relevant only if using LDAP
+                'new_ldap_ca_key': ''    # Relevant only if using LDAP
```
geokala, line 122: Do we need to support this and the prometheus ca key? I was under the impression that the only ones we actually do anything with (as part of the wait for start script) are the internal and external?

If we do need these then let's add a trailing comma to remove the misleading -1.

cloudify_cli/commands/certificates.py
```diff
@@ -113,14 +113,19 @@ def _get_configuration_dict(client):
...
                  },
             'postgresql_server': {  # Relevant only if ssl_enabled==True
                 'new_postgresql_server_cert': '',
                 'new_postgresql_server_key': '',
-                'new_postgresql_server_ca_cert': ''
+                'new_postgresql_server_ca_cert': '',
+                'new_postgresql_server_ca_key': ''
```
geokala, line 128: DItto for this one.

geokala (approved):

PR #1326, opened 2021-09-06
RD-3032 cfy audit_log list command implementation

mateumann:


cloudify_cli/commands/audit_log.py
```diff
@@ -0,0 +1,51 @@
...
+@click.option('-e', '--execution-id',
+              help=helptexts.AUDIT_EXECUTION_ID)
+@cfy.options.sort_by()
+@cfy.options.descending
+@cfy.options.pagination_offset
+@cfy.options.pagination_size
+@cfy.options.common_options
+@cfy.pass_logger
+@cfy.pass_client()
+def get(creator_name,
```
geokala, line 30: Shouldn't this be list rather than get?

cloudify_cli/table.py
```diff
@@ -62,6 +62,8 @@ def get_values_per_column(column, row_data):
                 row_data[column] = ','.join(row_data[column])
             elif isinstance(row_data[column], bool):
                 pass  # Taking care of False (otherwise would be changed to '')
+            elif isinstance(row_data[column], int):
+                pass  # Taking care of zero (otherwise would be changed to '')
```
geokala, line 66: This is a behaviour change- I think the tests should detect if anything breaks though.

mateumann: Yes, sure, that's something "big".  However that's an improvement IMO, so I didn't hesitate much :)

cloudify_cli/commands/audit_log.py
mateumann: Here are my reasons:
* I believe we won't ever need (I already know how it sounds…) to get a single audit log entry,
* I don't like to shadow `list()` function (or any other for that matter) unless I absolutely have to,
* it's a REST's GET after.

cloudify_cli/table.py
tehasdf: it's also a display-only change, and tests of course cannot say if the output looks better or worse ;)

cloudify_cli/commands/audit_log.py
```diff
@@ -0,0 +1,51 @@
...
+        creator_name=creator_name,
+        execution_id=execution_id,
+        order_by=sort_by,
+        desc=descending,
+        offset=pagination_offset,
+        size=pagination_size
+    )
+    print_data(AUDITLOG_COLUMNS, logs, 'AuditLogs:')
+    logger.info('Showing {0} of {1} nodes'
+                .format(len(logs), logs.metadata.pagination.total))
```
tehasdf, line 51:
```suggestion
    logger.info('Showing %s of %s nodes',
                len(logs), logs.metadata.pagination.total)
```

cloudify_cli/commands/audit_log.py
```diff
@@ -0,0 +1,51 @@
...
+@cfy.pass_client()
+def get(creator_name,
+        execution_id,
+        sort_by,
+        descending,
+        pagination_offset,
+        pagination_size,
+        logger,
+        client
+        ):
```
tehasdf, line 38: cmon if you put the closing paren on a separate line then a trailing comma on the last line is kinda a must have :P
```suggestion
        client,
        ):
```

cloudify_cli/commands/audit_log.py
tehasdf: how about list_logs, then

cloudify_cli/table.py
```diff
@@ -142,7 +144,7 @@ def print_details(data, title):
     """Utility for printing structured key/value pairs.
 
     Note that this is not for printing the standard output table, but
-    rather for auxilliary data.
+    rather for auxiliary data.
```
tehasdf, line 147: oh huh, I really thought it was spelled with a double-L!

cloudify_cli/commands/audit_log.py
```diff
@@ -0,0 +1,51 @@
...
+    """List audit_log entries"""
+    logger.info('Listing audit logs...')
```
tehasdf, line 40:
```suggestion
    logger.info('Listing audit log entries...')
```
Maybe this form? You use "entries" in both the docstring of this, and in the docstring of `auditlogs`

cloudify_cli/commands/audit_log.py
```diff
@@ -0,0 +1,51 @@
...
+    logs = client.auditlog.list(
+        creator_name=creator_name,
+        execution_id=execution_id,
+        order_by=sort_by,
+        desc=descending,
+        offset=pagination_offset,
+        size=pagination_size
+    )
+    print_data(AUDITLOG_COLUMNS, logs, 'AuditLogs:')
+    logger.info('Showing {0} of {1} nodes'
```
tehasdf, line 50: not `nodes` :)

cloudify_cli/commands/audit_log.py
```diff
@@ -0,0 +1,51 @@
+import click
+
+from ..cli import cfy, helptexts
+from ..table import print_data
+
+AUDITLOG_COLUMNS = ['ref_table', 'ref_id', 'operation', 'creator_name',
+                    'execution_id', 'created_at']
+
+
+@cfy.group(name='auditlogs')
```
tehasdf, line 10: this `name` is the user-visible side, and in this, I think we should go for `auditlog` singular. In principle, it's a log, with log lines, not multiple logs, right?

(in practice, we'll all just use `cfy au li` anyway)

tehasdf (commented):
I didn't have anything interesting to say, so I picked on the wording in a few places. Woo.

cloudify_cli/commands/audit_log.py
geokala: Seconding @tehasdf 's suggestion- whatever we do let's not call this get, since in the CLI that has semantics of "One thing".

cloudify_cli/table.py
geokala: Agreed, I'll mark this as resolved. Just wanted to make sure we were all approaching the coming testpocalypse with open eyes!

tehasdf (approved):

geokala (approved):

PR #1329, opened 2021-09-10
RD-3156 Wait for the upload_blueprint execution creation

mateumann:
Wait for upload_blueprint execution creation for maximum of about 5
seconds (0ms -> 100ms -> 200ms -> ... -> 900ms).  Sometimes the
execution was not started yet at the time we tried retrieving it's ID.

### DIFF
--- a/cloudify_cli/utils.py
+++ b/cloudify_cli/utils.py
@@ -34,6 +34,7 @@ from backports.shutil_get_terminal_size import get_terminal_size
 
 import yaml
 import requests
+from retrying import retry
 
 from .logger import get_logger, get_events_logger
 from .exceptions import CloudifyCliError, CloudifyTimeoutError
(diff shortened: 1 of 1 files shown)


cloudify_cli/utils.py
```diff
@@ -425,7 +426,21 @@ def _handle_errors():
...
+        bp = client.blueprints.get(blueprint_id)
+        # upload_execution['id'] might not be available at first, hence retry
+        return bp, bp.upload_execution['id']
+
+    try:
+        blueprint, execution_id = _get_blueprint_and_upload_execution_id()
+    except KeyError:
+        raise RuntimeError(
+            'Failed to get upload_blueprint workflow execution for blueprint '
+            '{0}.  That may indicate a problem with blueprint upload.  Verify '
```
geokala, line 440: Can this sometimes indicate a slow manager, or is it always some other problem?

tehasdf (approved):

cloudify_cli/utils.py
tehasdf: I think currently it can, but soon it will not anymore

geokala: OK, then I'm happy with this.

geokala (approved):

cloudify_cli/utils.py
mateumann: There could me multiple reasons for a failure, the most basic one would be a break in communication between mgmtworker and restservice.

PR #1333, opened 2021-09-27
RD-3252 Follow audit logs stream

mateumann:


cloudify_cli/commands/audit_log.py
```diff
@@ -31,18 +37,48 @@ def auditlog():
...
               pagination_offset,
               pagination_size,
               logger,
               client,
               ):
+    since_timestamp = before_to_utc_timestamp(since) if since else None
+    if follow:
+        raise CloudifyCliError('Streaming not available older Python versions.'
+                               '  Consider upgrading your installation to one '
+                               'based on Python>=3.6.')
```
tehasdf, line 54:
```suggestion
        raise CloudifyCliError('Streaming requires Python>=3.6.')
```
look how much shorter it is! :)
but the only reason I touched this, is the "consider". Don't be so soft on users, they have to be beaten into submission, not nicely asked to consider stuff! :P

cloudify_cli/async_commands/audit_log.py
```diff
@@ -0,0 +1,175 @@
...
+                  creator_name,
+                  execution_id,
+                  logger,
+                  client
+                  ):
+    """Truncate audit_log entries"""
+    before_timestamp = before_to_utc_timestamp(before)
+    if before_timestamp is None:
+        raise CloudifyCliError('Failed to parse timestamp: {0}'
+                               .format(before))
```
tehasdf, line 166:
```suggestion
        raise CloudifyCliError(f'Failed to parse timestamp: {before}')
```
you already use fstrings in this module

cloudify_cli/async_commands/audit_log.py
```diff
@@ -0,0 +1,175 @@
...
+async def _stream_logs(creator_name,
+                       execution_id,
+                       since,
+                       timeout,
+                       logger,
+                       client):
+    if not hasattr(client.auditlog, 'stream'):
+        raise CloudifyCliError('Streaming endpoint client not available. '
+                               'Consider upgrading your installation to one '
+                               'based on Python>=3.6.')
```
tehasdf, line 114:
```suggestion
        raise CloudifyCliError('Streaming requires Python>=3.6.')
```

cloudify_cli/async_commands/audit_log.py
```diff
@@ -0,0 +1,175 @@
...
+              pagination_size,
+              logger,
+              client,
+              ):
+    since_timestamp = before_to_utc_timestamp(since) if since else None
+    if follow:
+        loop = asyncio.get_event_loop()
+        if not loop:
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
```
tehasdf, line 59: I don't think this is necessary? shouldn't .get_event_loop() just create one?

cloudify_cli/__init__.py
```diff
@@ -13,3 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ############
+
+import sys
+PY2 = sys.version_info[0] == 2
```
tehasdf, line 18: you can also import this from `cloudify._compat`. Using that c-common one is the reason c-cli never grew a `_compat.py` itself

cloudify_cli/async_commands/audit_log.py
mateumann: Oh, in deed it does

tehasdf (dismissed):

cloudify_cli/async_commands/audit_log.py
```diff
@@ -0,0 +1,169 @@
...
+              follow,
+              timeout,
+              sort_by,
+              descending,
+              pagination_offset,
+              pagination_size,
+              logger,
+              client,
+              ):
+    since_timestamp = before_to_utc_timestamp(since) if since else None
```
geokala, line 54: Perhaps the function should just be called to_utc_timestamp now?

cloudify_cli/commands/audit_log.py
```diff
@@ -22,6 +22,12 @@ def auditlog():
               help=helptexts.AUDIT_CREATOR_NAME)
 @click.option('-e', '--execution-id',
               help=helptexts.AUDIT_EXECUTION_ID)
+@click.option('-i', '--since',
+              help=helptexts.AUDIT_SINCE)
+@click.option('-f', '--follow', '--tail',
```
geokala, line 27: Can we maybe not support --tail for this so that we can gradually phase out that wrong argument naming?

geokala: (I mean not support that as the name of the argument- just calling it --follow is correct)

cloudify_cli/commands/audit_log.py
```diff
@@ -31,18 +37,46 @@ def auditlog():
...
               sort_by,
               descending,
               pagination_offset,
               pagination_size,
               logger,
               client,
               ):
+    since_timestamp = before_to_utc_timestamp(since) if since else None
+    if follow:
+        raise CloudifyCliError('Streaming requires Python>=3.6.')
```
geokala, line 52: Assuming this command gets overridden on 3.6+ by the asyncio one, perhaps this should just not accept the follow arg, and instead have a comment in its help indicating that python should be upgraded to get more functionality?

cloudify_cli/main.py
```diff
@@ -48,7 +50,10 @@
 from .commands import certificates
 from .commands import node_instances
 from .commands import maintenance_mode
-from .commands import audit_log
+if PY2:
+    from .commands import audit_log
+else:
+    from .async_commands import audit_log
```
geokala, line 56: Is there a reason you didn't put the async form into the existing commands.audit_log and gate behaviour on available features?

tehasdf: py2 must not even attempt to import a module with those functions, because being `async def`, using `async for`, they'll be a syntaxerror

cloudify_cli/commands/audit_log.py
tehasdf: this is more visible though. I can imagine users calling support to ask "why don't I have streaming support???", because they didn't read the helptxt. But this is impossible to miss.

geokala: Fair enough- it just looked a bit strange at first glance, but makes sense.

cloudify_cli/main.py
geokala: There's only one function in there (and not the command itself) that uses those keywords. Farming them out to a separate module and importing that only if $correct_py_ver would result in less repitition on defining the commands, etc.

cloudify_cli/async_commands/audit_log.py
mateumann: good point, thx

cloudify_cli/commands/audit_log.py
mateumann: sure thing, I wasn't sure if now was the good time for that, I'll gladly drop `--tail` from the list

geokala (approved):

PR #1334, opened 2021-10-05
local: allow listing blueprints even if there's none

tehasdf:
I just did `cfy prof use local` and `cfy blu li` and was greeted with
an error. I expected to see an empty list instead.
So this makes it be an empty blueprints list indeed.

### DIFF
--- a/cloudify_cli/cli/cfy.py
+++ b/cloudify_cli/cli/cfy.py
@@ -1776,6 +1776,16 @@ class Options(object):
             f = arg(f)
         return f
 
+    def local_common_options(self, f):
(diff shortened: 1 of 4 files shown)


cloudify_cli/cli/cfy.py
```diff
@@ -1776,6 +1776,16 @@ def common_options(self, f):
             f = arg(f)
         return f
 
+    def local_common_options(self, f):
```
tehasdf, line 1779: the main point of this is that of course I'm going to be using it for many local commands soon.
But for now, just the one.

geokala (approved):

PR #1349, opened 2021-12-01
RD-3630 cfy blueprints set-owner implementation

mateumann:


cloudify_cli/commands/blueprints.py
```diff
@@ -580,6 +580,20 @@ def set_icon(blueprint_id, icon_path, logger, client):
...
+@cfy.argument('blueprint-id')
+@cfy.options.chown_username()
+@cfy.assert_manager_active()
+@cfy.pass_client()
+@cfy.pass_logger
+def chown(blueprint_id, username, logger, client):
+    """Set a new owner for the blueprint."""
+    client.blueprints.update(blueprint_id, {'creator': username})
+    logger.info('Blueprint `{0}` is now owned by user `{1}`.'
+                .format(blueprint_id, username))
```
tehasdf, line 594:
```suggestion
    logger.info('Blueprint `%s` is now owned by user `5s`.',
                blueprint_id, username)
```
let's keep using this convention i think

tehasdf (commented):
I seem to remember we agreed on the `set-owner` user rather than `chown`? (for similarity with all the other `set-*` subcommands)

tehasdf:
it was set-owner indeed! see RD-3239

tehasdf (approved):

PR #1365, opened 2021-12-28
RD-3900 Cope with deployments in local workflows

tehasdf:
I'm adding the concept of deployments to local workflows,
so the CLI must at least cope with that.

This doesn't add any new features to the CLI, yet, and this
is not the "cfy local fix" just yet.

This just allows keeping old cfy local still working mostly the
same way as it was before.


geokala (approved):

tehasdf:
force-push was dropping dev-reqs

geokala (approved):

PR #1373, opened 2022-01-14
RD-3985 cfy blueprints get labels fix

mateumann:
`cfy blueprints get` failed for blueprints which had some labels defined.  This patch fixes that.

And also let's display information about plugins this blueprint has in its plan (because it's easy and can be helpful).

### DIFF
--- a/cloudify_cli/commands/blueprints.py
+++ b/cloudify_cli/commands/blueprints.py
@@ -306,6 +306,13 @@ def get(blueprint_id, logger, client, tenant_name):
     blueprint_dict['#deployments'] = len(deployments)
     columns = BLUEPRINT_COLUMNS + ['#deployments']
     blueprint_metadata = blueprint_dict['plan']['metadata'] or {}
+    blueprint_plugins = {k: [p for p in blueprint_dict['plan'][k]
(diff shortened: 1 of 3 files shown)


cloudify_cli/commands/blueprints.py
```diff
@@ -315,6 +315,7 @@ def get(blueprint_id, logger, client, tenant_name):
         blueprint_dict['deployments'] = blueprint_deployments
         print_single(columns, blueprint_dict, 'Blueprint:', max_width=50)
     else:
+        modify_resource_labels([blueprint_dict])
```
geokala, line 325: What does modify_resource_labels actually do? It sounds like it changes things before we display them to the user, which doesn't sound like something we should be doing. Is it just badly named?

mateumann: It's just serializing labels: the function we use in `cfy blu list` and `cfy dep list` (https://github.com/cloudify-cosmo/cloudify-cli/blob/1a71412c50b9ff1a35f366008133f74326ce0fe1/cloudify_cli/labels_utils.py#L8-L17)

geokala (approved):

PR #1376, opened 2022-02-06
Bump PyYaml version to 6.0 only for py3.6+

glukhman:


setup.py
```diff
@@ -16,6 +16,32 @@
...
 
+install_requires = [
+        'click>7,<8',
+        'wagon[venv]>=0.10.1',
+        'jinja2>=2.10,<2.11',
+        'retrying==1.3.3',
+        'colorama==0.4.4',
+        'requests>=2.7.0,<3.0.0',
+        'click_didyoumean==0.0.3',
+        'cloudify-common[dispatcher]==6.3.0.dev1',
```
yuvlar, line 27: I believe this should be 6.4.0.dev1

setup.py
```diff
@@ -16,6 +16,32 @@
 
 from setuptools import setup
 
+install_requires = [
+        'click>7,<8',
+        'wagon[venv]>=0.10.1',
+        'jinja2>=2.10,<2.11',
```
yuvlar, line 22: jinja2 should be pinned to 2.11.3

yuvlar (approved):

PR #1384, opened 2022-03-23
RD-4344 Catch all manager errors when retrieving version info

glukhman:


cloudify_cli/cli/cfy.py
```diff
@@ -167,10 +167,14 @@ def show_version(ctx, param, value):
...
         'version': pkg_resources.require('cloudify')[0].version
     }
-    rest_version_data = env.get_manager_version_data() \
-        if env.is_manager_active() else None
+    try:
+        rest_version_data = env.get_manager_version_data() \
+            if env.is_manager_active() else ""
+    except Exception as e:
+        rest_version_data = "Cloudify Manager unreachable. {}: {}\n".format(
+            type(e).__name__, str(e))
```
geokala, line 175: Why not just output it to stderr at this point and exit with status 1?

glukhman: We need to output the CLI version in any case, so we can't exit quite yet here

geokala: The problem is that this is then giving an exit status of 0 for a command that failed, and not putting the output in the right place (stderr for a problem).

tehasdf (approved):

PR #1387, opened 2022-03-24
Import ABC from collections.abc for Python 3.10 compatibility.

tirkarthi:
Use `collections` for Python 2 and `collections.abc` for Python 3.

### DIFF
--- a/cloudify_cli/utils.py
+++ b/cloudify_cli/utils.py
@@ -27,10 +27,14 @@ import logging
 import tarfile
 import zipfile
 import tempfile
-import collections
 from shutil import copy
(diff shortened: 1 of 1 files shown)


geokala:
Thanks for your compat fix, @tirkarthi - much appreciated!

geokala (approved):

tehasdf:
I Just can't get jenkins to deal with this :(
@yuvlar there's no way for us to run CI on non-`cloudify-cosmo` remotes

Therefore, I've copied the commit from this PR into #1420 and merged it there. Thanks!

PR #1389, opened 2022-03-25
Make use not throw an error due to missing profile

geokala:


geokala:
This also means we're now constructing the profile entirely in memory, rather than partially constructing it then fiddling around, saving it to disk, doing a bit more, then finishing constructing it and saving it again.

tehasdf (approved):

PR #1391, opened 2022-03-29
Don't check auth before setting profile settings 

geokala:
We check before saving anyway, and checking before processing
the new settings will run into interesting issues (e.g. when
trying to set a token while username and password are already set).


geokala:
The place we currently check is https://github.com/cloudify-cosmo/cloudify-cli/pull/1391/files#diff-e87cfbb14148a939dd1e0d4b880872fc7204e85b7b04b0d610dcef390d4dbc07R433

tehasdf (approved):

PR #1408, opened 2022-05-24
RD-4355 attempt to build with pyoxidizer

mateumann:


mateumann:
we've decided to use Nuitka https://github.com/cloudify-cosmo/cloudify-cli/pull/1396 + https://github.com/cloudify-cosmo/cloudify-build-system/pull/654

PR #1419, opened 2022-06-23
RD-5128 Fix `--all-tenants` functionality in plugins update

mateumann:
Just call the operation in the blueprint tenant's context

### DIFF
--- a/cloudify_cli/commands/plugins.py
+++ b/cloudify_cli/commands/plugins.py
@@ -562,20 +562,22 @@ def update(blueprint_id,
             'only the specific plugins, use --plugin-name parameter instead.')
 
     if blueprint_id:
-        _update_a_blueprint(blueprint_id,
-                            plugin_names,
-                            to_latest,
-                            all_to_latest,
-                            to_minor,
-                            all_to_minor,
-                            include_logs,
-                            json_output,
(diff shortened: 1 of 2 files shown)


cloudify_cli/commands/plugins.py
```diff
@@ -560,32 +561,48 @@ def update(blueprint_id,
...
+                            to_minor,
+                            all_to_minor,
+                            include_logs,
+                            json_output,
+                            logger,
+                            force,
+                            auto_correct_types,
+                            reevaluate_active_statuses,
+                            client,
+                            tenant_name)
```
tehasdf, line 578:
```suggestion
        _update_a_blueprint(
            blueprint_id,
            plugin_names,
            to_latest,
            all_to_latest,
            to_minor,
            all_to_minor,
            include_logs,
            json_output,
            logger,
            force,
            auto_correct_types,
            reevaluate_active_statuses,
            client,
            tenant_name,
        )
```
I suggest this style: you say you want the arguments to be one-line-each, so why not go even one step further, and have indeed _all_ the args one-line-each, even the last one: with the trailing comma and the `)` on the separate line

geokala (approved):

PR #1425, opened 2022-07-08
RD-4466 dep-up: add the force-reinstall flag

tehasdf:
This will make the changed nodes be just reinstalled, bypassing
the update interfaces


tehasdf:
Related to cloudify-cosmo/cloudify-common#1109

geokala (approved):

PR #1438, opened 2022-09-09
Remove unused libraries from setup.py

tehasdf:
Things that the cli doesn't DIRECTLY depend on, don't go in setup.py

Also no reason to pin wheel.
(no, Pinwheel is a Dark Souls 1 boss).


geokala (approved):
Agreed, we don't want a soulslike CLI.

PR #1444, opened 2022-10-24
RD-6110 Add community register command + skip license check for uploa…

glukhman:
…d commands on community


cloudify_cli/commands/community.py
```diff
@@ -0,0 +1,39 @@
+import click
+
+from cloudify_cli.cli import cfy
+
+
+@cfy.group(name='blueprints')
```
tehasdf, line 6: whoops

tehasdf (approved):

PR #1448, opened 2022-11-14
RD-6061 Provide get, update and delete commands for Secrets Providers

ghost:


cloudify_cli/commands/secrets.py
```diff
@@ -397,6 +399,112 @@ def provider_create(
     )
 
 
+@providers.command(
+    name='update',
+    short_help='Update a already exists secrets provider',
```
tehasdf, line 404:
```suggestion
    short_help='Update an existing secrets provider',
```

cloudify_cli/cli/cfy.py
```diff
@@ -2450,6 +2434,35 @@ def blueprint_filter_rules(self, f):
...
+            'help': _help or helptexts.SECRET_PROVIDER_TYPE,
+            'callback': validate_value_not_empty,
+        }
+
+        return click.option(*args, **kwargs)
+
+    @staticmethod
+    def connection_parameters(required=True, _help=None, default=None):
+        args = [
+            '--connection_parameters',
```
tehasdf, line 2454: eh maybe let's make this `--connection-parameters`? underscores in arguments are a bit weird i think

cloudify_cli/commands/secrets.py
```diff
@@ -397,6 +399,112 @@ def provider_create(
...
+)
+@cfy.options.secret_provider_name
```
tehasdf, line 406: I think `.secret_provider_name` doesn't exist? :)

Anyway, I think it'd be better to have the name (in update and in delete) be an argument rather than a flag, ie. `@cfy.argument('secret-provider-name')`.
So that it's `cfy sec prov del provider1` rather than `cfy sec prov del --name provider1`

tehasdf (dismissed):

tehasdf (approved):

